Issue #2, opened 2024-09-14
Overloading Creates Issues for Variables that aren't functions


HalflingHelper:
The Following program fails:
```
import Nat
define x = 7
define x = 8
print suc(x)
```
with the error 
```
test.pf:4.11-4.12: expected a term of type Nat
but got term x of type (x.8: Nat & x.1111: Nat)
```

This appears to be a problem caused by the type checker and how it handles overloading defines with repeated variable names, and being unable to figure out the situation when they are passed as function arguments.

This is a big issue in particular with the libraries. For example `List.pf` has `L1` defined with in, leading to any program intending to use `L1` as a name failing when it tries to pass it as a function argument (A 343 student did exactly this, which is what led me to pursue this bug).

I'm not sure what the best solution is
- Do we change the type checker to only call `add_overload` when the variable name is bound to a function type. Otherwise not overloading and just overriding? This feel like the most natural to me, but I'm not sure what the best way is to detect that with this architecture.
- Is there some way to detect overloads in `type_check_term`, and choose the appropriate one? This feels like sketchy behavior and not something that happens in any language I know of. This feels especially bad because we could write programs like:
```
import Nat
define x = 7
define x = true
print suc(x)
```
which have two values of `x` floating around with different types 🙃.

HalflingHelper:
> I'm not sure what the best way is to detect that with this architecture.

Having said this I modified the `Define` case of `process_declaration` in a way that appears to solve the issue (passing all tests).

I'm happy to add additional tests for the cases that cause the bug if you think this is a good solution.

https://github.com/jsiek/deduce/pull/3

PR #3, opened 2024-09-14
Only overloading functional values

HalflingHelper:
See https://github.com/jsiek/deduce/issues/2


jsiek (approved):
I like this change. Indeed, overloading is only useful for functions.

We'll also need to add a way to hide stuff (private) so things like L1 don't get exported, but that's a separate issue.

PR #6, opened 2024-09-21
Add verbose command line arg

Temperz87:
Added "--verbose" to print verbosely
Removed some if statements and call setters directly


jsiek (approved):
Very nice!

PR #7, opened 2024-09-22
Adding a test for redefinition of variables

HalflingHelper:


HalflingHelper:
Just adding this so that we can close https://github.com/jsiek/deduce/issues/2

PR #8, opened 2024-09-22
Gooder paths

Temperz87:
Less hardcoding because It was hurting my brain

I highly encourage @jsiek to make sure the Makefile still works before merging, mainly making sure that it's correctly finding python as it's not in your path.


jsiek:
Would you please write a couple sentences that summarize the changes?

Temperz87:
The changes all concern the Makefile, in that a lot of paths were hardcoded, and if I so muhc as renamed a file everything would break.
$(PYTHON) used to be at "/opt/homebrew/bin/python3.10", however it's now determined by "$(shell command -v python3)", which should automatically locate Python for us, instead of having to update the Makefile when we clone the repository, or switch branches, and then reupdating before pushing. This change is my primary reason for wanting you to test out the file before merging it in, as I want to make sure that the command will work on Mac.
All tests used to be in a directory called "test", and in the Makefile there were tons of lines such as `$(PYTHON) ./deduce.py test/theorem_true.p`, however now every file is either in the root directory of the repository, in "test/should-pass", or "test/should-error", and are run through a for loop in the Makefile.

jsiek (approved):
These changes are great, thank you!
I've confirmed that this works on my machine.
(After installing lark in the python that gets found by the Makefile.)

PR #10, opened 2024-09-26
Change how command lines arguments get parsed

Temperz87:
Move the behavior associated with parsing and checking a file into a function named "deduce_file"
Make file to parse not expected to be the first argument passed in
Add TODO message for when multiple files are passed in


jsiek (approved):
Looks great, thank you!

PR #13, opened 2024-09-30
Add induction advice alongside arbitraries

Temperz87:
Mainly, add induction device (wowiee!!!) to appear alongside advice for arbitrary variables, but not types. The advice follows the recipe of "If that fails, you can try induction with" followed by copy and pastable induction cases with proof holes inside of them. 

One thing that's a bit weird is the naming system, which starts at line 682 of proof_checker.py, where I have predefined names (p, q, r, z, y, x) that with every parameter passed into a constructor, become "x'" "xs'" "xss'" etc. etc. This was the best solution I could think of, but I'm sure that someone else can do better.

Last note: if a union isn't recursively defined, it'll still print out induction advice, maybe we should change that?


jsiek (approved):
Excellent!

PR #16, opened 2024-10-05
If and Only If

HalflingHelper:
Updated the parser to handle iff, implemented iff, added tests, and started rewriting the theorems in `Nat.pf` that can be rewritten with the help of iff.

Bonus: Fixed some tiny issues with the suggestion for `arbitrary`


jsiek (commented):
This is great!

One question: does this impact the error message regarding when someone incorrectly uses modus ponens? For example, when there is a mismatch between the premise and the fact being suppied? Is the error message the same, better, or worse then before?

jsiek:
If it turns out the error message got worse, then we should consider creating a new syntactic form for using (eliminating) a biconditional instead of overloading the `apply`-`to` syntax.

HalflingHelper:
I just pushed some commits this morning, so maybe they've gotten better for this PR.

The error message should be the same for `apply (if p then q) to p`, since I didn't change the behavior there.

When applying a forall, the message will be the same if the program can't deduce an instantiation for the variables.

When working with and, there are two parts where things can go wrong. One is when doing the application, and the other is checking that what you get actually implies the target. The latter will use the normal and error, the other when we can't actually apply any of the conjuncts.

I added two messages for this, one when none of the conjuncts are if-then, and another for when they can't be applied, those being
```
theorem test1 : 1 <= 2
proof
  have step1 : 1 * 2 = 2 by one_mult[2]
  have step2 : 1 * 3 = 3 by one_mult[3]
  have step3 : 1 * 2 = 2 and  1 * 3 = 3 by step1, step2
  have step4 : true by apply step3 to step2
  ?
end
-------
iff.pf:93.24-93.44: in 'apply', expected at least one if-then formula as a conjunct of (1 * 2 = 2 and 1 * 3 = 3)
```
and
```
theorem test: 1 <= 2
proof
    have z_l_one: 0 <= 1 by definition { operator <=}
    have step2 : (1 <= 2 <=> 2 <= 3) by suc_less_less_suc[1, 2] 
    have t3 : true by apply step2 to z_l_one
    ?
end
--------------
iff.pf:106.23-106.45: could not prove that 0 ≤ 1 implies at least one of
        1 ≤ 2
        2 ≤ 3
for application of 
        (1 ≤ 2 ⇔ 2 ≤ 3)
to 
        z_l_one
```

tl;dr - Errors should be the same, with added cases for when and runs into issues. The only thing I could maybe see as a problem is that when trying to apply something like (all x: Nat. P), the error will be "could not deduce an instantiation" if there is a problem in proving any sort of implication. For example this error can happen:
```
iff.pf:97.24-97.47: could not deduce an instantiation for any of the variables for application of 
        (all x:Nat, y:Nat. (if x ≤ y then suc(x) ≤ suc(y)))
to
        step2
```
where step 2 is `have step2 : all x: Nat. x * 1 = x by mult_one`.
The current error for this is:
```
iff.pf:96.24-96.47: formula: (all y:Nat, x:Nat. x * 1 = x)
does not match expected formula: suc(x) ≤ suc(y)
while trying to deduce instantiation of
        (all x:Nat, y:Nat. (if suc(x) ≤ suc(y) then x ≤ y))
to apply to
        (all y:Nat, x:Nat. x * 1 = x)
```
The reason for this problem is that when deducing the instantiation, there can be more than one (prem, conc) pair, and we only need one of them to match. For example `apply (all x:Nat. x > 0 <=> suc(x) > 1) to (y > 0)`, only the rightward direction will have an instantiation.

A solution I can maybe think about is maybe saving the errors in the All case, and then if something fails printing them all out, as a dump of -> here was what we tried for each conjunct, and here is what went wrong in detail, unless you're ok with the relatively vague "could not deduce an instantiation" (which is technically true 🙃)

jsiek:
Thanks for the thorough answer! This is good enough for now and we can do improvements to the All case as a later PR.

PR #19, opened 2024-10-08
Make deduce able to process directories

Temperz87:
When you pass in a directory to deduce, it now looks for all files ending in ".pf" and processes them. If the "-r" or "--recursive-directories" argument is detected, it'll also delve into sub directories.


jsiek (approved):
This looks great, thanks!

Issue #14, opened 2024-10-01
Define Syntax Sugar for List Construction


mateidragony:
Similar to how deduce parser handles digits where:
```
3 -> suc(suc(suc zero)))
```
We could do a thing where a "nicer" syntax for lists could be implemented:
```
[1, 2, 3] -> Node(1, Node(2, Node(3, empty)))
```
Just a thought 😁

PR #20, opened 2024-10-08
Define syntax sugar for list construction

mateidragony:
Add List syntax sugar according to #14 

Added List unwrapping from parser for:
- [] -> empty
- [1, ...] -> node(1, ...)

Added List rewrapping for printing and error messages:
- empty -> []
- node(1, ...) -> [1, ...]

One thing I had a question on is why I had to pattern match my node calls with `TermInst` instead of `Var` like what `natToInt` and `intToNat` does
- abstract_syntax.py:2344 and abstract_syntax.py:2346
```python
case TermInst(loc2, tyof2, Var(loc3, tyof3, name), tyargs, True) if base_name(name) == 'empty':
    return True
case Call(loc, tyof1, TermInst(loc2, tyof2, Var(loc3, tyof3, name), tyargs, True), [arg, ls], infix) if base_name(name) == 'node':
    return isNodeList(ls) 
```
- abstract_syntax.py:2353 and abstract_syntax.py:2355
```python
case TermInst(loc2, tyof2, Var(loc3, tyof3, name), tyargs, True) if base_name(name) == 'empty':
    return ''
case Call(loc, tyof1, TermInst(loc2, tyof2, Var(loc3, tyof3, name), tyargs, True), [arg, ls], infix) if base_name(name) == 'node':
    return str(arg) + ', ' + nodeListToList(ls)
```

I took a look at the `type_check_term` function in `proof_checker.py` and I think it's something from there, but I'm not sure what that difference implies and if it's a difference that we should concern ourselves with. 

I think this occurs somewhere around here: https://github.com/jsiek/deduce/blob/035f76a67750d97c4dff6d4e62af7300657a43e4/proof_checker.py#L1770-L1800


jsiek:
Regarding TermInst, the List union is generic (has a type parameter) whereas Nat is not generic. When a union is generic, so are its constructors. So TermInst is needed to instantiate the constructor, willing in a type argument for the type parameter.

jsiek:
Please add some tests. Thanks!

mateidragony:
I added some tests for just regular functional programming and proof writing. Also I found an error which I fixed...

jsiek (approved):
This looks great, thank you!

PR #22, opened 2024-10-10
Cj/sub thms

HalflingHelper:
I wanted to prove sub_dist_mult, and in doing so had to create some other theorems. They all seem to match the vibe of other theorems proved in `Nat.pf`, but feel free to nix any that don't seem relevant.


jsiek:
Excellent, thanks!

PR #24, opened 2024-10-24
Integers!!!

mateidragony:
We currently have import errors....

- Added Int.pf library file which includes:
  - The int union type
  - negation function for Int and Nat -> Int
  - Addition, subtraction, and multiplication between (Int,Int), (Int,Nat), and (Nat,Int) -> Int
- Added int_error_for_now.pf file in the should_error tests directory
  - This file contains all of my current tests for integers
  - It currently errors out due to our overloading import errors

### DIFF
--- a/lib/Int.pf
+++ b/lib/Int.pf
@@ -1,0 +1,70 @@
+import Nat
+
+union Int{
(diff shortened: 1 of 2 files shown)


jsiek:
Great!

PR #29, opened 2024-10-31
Gooder testing

Temperz87:
Added `test-deduce.py` with the hope of replacing the Makefile. When run with zero arguments, this file with test all targets with both parsers, but there are specific arguments to target certain folders.

`--regenerate-errors` creates .pf.err files
`--generate-errors FILE` generates errors for a certain file
`--lib` will  only test the lib
`--passable` will only test the test-should-pass directory
`--errors` will only test the test-should-error directory

Note: you can have multiple "targets" to test, e.g. "py test-deduce.py --passable --lib"


jsiek:
Awesome, thanks!

PR #30, opened 2024-10-31
I forgot to fix this in the testing script

Temperz87:
The testing script now detects python correctly
Whoopsies!!!!!!


jsiek:
what is `actual_error.tmp` for?

Temperz87:
Oh no that should've also been deleted. Basically the testing script dumps the output of running deduce into a file then uses diff to check the error messages, however if you ctrl+c the program sometimes that file doesn't get deleted. It should be fine to keep as it'll be gone the next time the tester gets run

Issue #27, opened 2024-10-31
Error Messages Unclear wrt Theorem type instantiation


HalflingHelper:
```
import List
import Nat

theorem l_one : length(node(1, empty)) = 1
proof
  // length_one[1]
  // length_one[Nat][1]
  // length_one[Nat, 1]
  length_one<Nat>[1]
end
```

Error messages here for the commented errors are unclear as to what the problem is. Working on fixing this.

Ex In Order: 
`expected type but the call returns Nat`

```
/home/calvin/Documents/Programming/343/deduce/test.pf:6.3-6.18: to instantiate:
        length_one : (all U:type. (all x:U. length([x]) = 1))
with type arguments, instead write:
        length_one<Nat>
```

```
/home/calvin/Documents/Programming/343/deduce/test.pf:6.3-6.21: to instantiate:
        length_one : (all U:type. (all x:U. length([x]) = 1))
with type arguments, instead write:
        length_one<Nat, 1>
```

PR #31, opened 2024-11-05
All Simplification and Errors

HalflingHelper:
## Changes
- Commas in `all`, `arbitrary`, and theorem instantiation are now syntactic sugar
  - Removed lists from AST nodes, and added position property to allow for pretty printing.
- Error message suggesting a type argument when someone forgets it (https://github.com/jsiek/deduce/issues/27)

## Possible Enhancements
- [] Factor out repeated code for wrapping and desugaring
  - I want to make sure that it's still clear what the code is doing
  - I didn't see anywhere obvious to add utility functions

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-PYTHON = $(shell command -v python3.11)
+PYTHON = $(shell command -v python3.12)
 
 LIB_DIR = ./lib
 TEST_PASS_DIR = ./test/should-pass
--- a/abstract_syntax.py
+++ b/abstract_syntax.py
@@ -1395,7 +1395,7 @@ class All(Formula):
     result = ''
     if s + 1 == e:
       result += "(all "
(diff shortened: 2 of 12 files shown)


jsiek:
Calvin, this is great!
Have you run `test-deduce.py` to check that all the tests still pass?
Did you add any new tests or change the errors that arise from old tests?

HalflingHelper:
Oh I just pulled in that addition from Shulin, looks like I have one or two small things to fix.

HalflingHelper:
Some of these should error also seem to have sub-optimal error messages caused by private imports thing, so I'm changing these to import Nat.

i.e.
```
./test/should-error/inst-type-square-bracket.pf:4.17-4.18: undefined variable `suc`	(uniquify)
```

jsiek:
ok, I'll wait for you to commit those further changes before I merge this PR

PR #33, opened 2024-11-12
All intro fix printing

HalflingHelper:
Fixed printing of goals / advice after an all intro by adding a method to strip the head of an All chain and update the positions of everything else


jsiek:
Thanks!

PR #37, opened 2024-11-17
Remove Marks Fix

HalflingHelper:
Removing marks from instantiation from `have`. Added a test `have_remove_marks`


jsiek (approved):
These changes look great, thank you!

PR #39, opened 2024-11-21
Make the tester automatically delete .thm files

Temperz87:
Add "Make clean", which will delete all .thm files in `lib/` and `tests/should-pass/`
Make `test-deduce.py` get rid of .thm files in in `lib/` and `tests/should-pass/`


jsiek:
Thanks!

Issue #42, opened 2024-12-16
More suffices advice stuff


HalflingHelper:
This goes back to the issue we discussed previously about suffices not showing the `@ <>` annotation.

The example below prompts:
```
/home/calvin/Documents/Programming/343/deduce/exercise.pf:17.5-17.33: 
suffices to prove:
        0 = foldr([], 0, +)
```
But I need to provide `0 = foldr(@[]<Nat>, 0, operator+) `. The operator thing is new and I don't see it being an issue for what is covered in the course (and maybe something fixed with a better parser anyway). It's weird though that the `TermInst` fix didn't work for this case...

```
// I'm doing a bunch of a exercises as a way to hopefully find
// things we can improve

import List
import Nat

function sum(List<Nat>) -> Nat {
  sum(empty) = 0
  sum(node(n, ns)) = n + sum(ns)
}


theorem foldr_eq_sum : all ls : List<Nat>. sum(ls) = foldr(ls, 0, operator+)
proof
  induction List<Nat>
  case empty {
    suffices ? by definition sum
    ?
  }
  case node(a, d) {
    ?
  }
end
```

jsiek:
Thank you! I'll look into this today.

jsiek:
I've fixed the issue with operator +.

jsiek:
The problem with `[]` is tricky... I think it's an interaction with the TermInst on foldr, which is causing the type checker to think it doesn't need to make the TermInst on `[]` explicit. But the TermInst on foldr doesn't get printed...

HalflingHelper:
I see you made a change that at least fixes the issue, event though it's less concise - would it be worth looking into adding additional logic to reduce, and changing inference there?

For example adding this case to `reduce()` in `TermInst` annotates vars only when we reduce in suffices. Could we do a bit more to only setting inferred to `False` ?
```
case Var(loc2, TypeInst(), name):
        return TermInst(self.location, self.typeof, subject_red,
                        self.type_args, False)
 ```

(This is still more verbose than necessary, for example when using `definition length` on the goal `length(node(1, node(1, empty)))` we get `1 + length(node(1, @empty<Nat>))`)

This doesn't change any of the existing error messages we test for though, except for the one you added for  this foldr example.

jsiek:
[...]

PR #44, opened 2024-12-18
Suffices terminst advice

HalflingHelper:
#42

### DIFF
--- a/abstract_syntax.py
+++ b/abstract_syntax.py
@@ -891,9 +891,26 @@ class Call(Term):
     return ret
 
   def substitute(self, sub):
-    ret = Call(self.location, self.typeof, self.rator.substitute(sub),
-                [arg.substitute(sub) for arg in self.args],
-                self.infix)
+    # TODO: factor out this internal helper function
+    def annot_terminsts(l):
+      match l:
+        case TermInst(loc, tyof, idk, type_args, inferred):
+          return TermInst(loc, tyof, idk, type_args, False)
+        case _:
+          return l
+        
+    fun = self.rator.substitute(sub)
+    match fun:
+      case TermInst():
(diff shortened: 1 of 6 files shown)


HalflingHelper:
Passing all tests, and I added two more where advice should / shouldn't be given. It might be worth taking some time to brainstorm more situations when this could arise, though as arguments of calls is certainly the most common.

HalflingHelper:
Wait to merge this, there might be a problem with the operator fix : 
```
suffices to prove:
        a + sum(d) = + a
```

HalflingHelper:
Also the lalr parser doesn't like this :)
theorem foldr_eq_sum : all ls : List<Nat>. sum(ls) = foldr(ls, 0, operator +)

HalflingHelper:
It's the paren after operator+

HalflingHelper:
Closing since https://github.com/jsiek/deduce/commit/0e88ca917e471eadf085805b71f9de83a0d760d7 appears to fix

jsiek:
Interesting. You're solution is a more specific that what I just pushed, which could make it less verbose, which is good, but the question is whether `Call` is the only problematic context...

jsiek:
I suspect that there are more contexts that could be problematic... any place in synthesis mode.

jsiek:
So the really good solution would be to augment the `substitute` method to know about check/synthesis and only set `inferred = False` when substituting in a synthesis context. But that's a bunch more work. I'll make an issue for that.

jsiek:
Hmm, actually, it's not as simple as recognizing synthesis mode because of generic functions.

jsiek:
I think perhaps the fix I just pushed is good because it is conservative and not that complicated. To do better seems to require a **much** more complex solution.

PR #47, opened 2024-12-30
Support configuration of library paths

cty12:
# Description

I added support for configuring library paths in `~/.config/deduce/libraries`. 

The format of the config file is similar to that of Agda (`~/.agda/libraries`): each line contains an absolute path to a library. 

For example, if I add the `lib/` directory to the config file: 

```
$ cat ~/.config/deduce/libraries
/Users/tianyu/workspace/deduce/lib
```

I can check a deduce source file that imports `Nat.pf` without using the `--dir` command line argument: 

```
$ cat test/should-pass/some1.pf
import Nat

theorem T: some x : Nat. x = x
proof
  choose 0
  reflexive
end%

$ python3 ./deduce.py test/should-pass/some1.pf
test/should-pass/some1.pf is valid
```

# Demo

This also works interactively in Emacs (I'm using the code of `deduce-mode` for which I submitted[ a pull request here](https://github.com/mateidragony/deduce-mode/pull/1)):

Screenshot 2024-12-30 at 13 34 51

# Testing

The code passed all test cases (`test-deduce.py`). I changed `test-deduce.py` to prioritize the latest Python version: for example, if I have Python 3.11 and 3.13 on my machine, the test script should use 3.13.

### DIFF
--- a/abstract_syntax.py
+++ b/abstract_syntax.py
@@ -26,15 +26,25 @@ def base_name(name):
     ls = name.split('.')
     return ls[0]
 
-import_directories = ["."]
+import_directories = set()
+
+def init_import_directories():
+  import_directories.add(".")
+  lib_config_path = Path(os.path.expanduser("~/.config/deduce/libraries"))
+  if lib_config_path.exists() and lib_config_path.is_file():
+    with open(lib_config_path, 'r') as lib_config_file:
+      for line in lib_config_file:
+        import_directories.add(line.strip())
   
 def get_import_directories():
   global import_directories
+  if(get_verbose()):
+    print("import directories: ", import_directories)
   return import_directories
 
 def add_import_directory(dir):
   global import_directories
-  import_directories.append(dir)
+  import_directories.add(dir)
 
 
 recursive_descent = True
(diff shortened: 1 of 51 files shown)


jsiek:
is there a way for you to check if lark is installed for a particular version of python? I might have lark installed with 3.11 but not 3.13

cty12:
> is there a way for you to check if lark is installed for a particular version of python? I might have lark installed with 3.11 but not 3.13

@jsiek Yes! You can use 
```sh
python3.13 -m pip list
```
To install lark with 3.13, try (remove `--user` for a global install)
```sh
python3.13 -m pip install --user lark
```

Temperz87:
I think Siek was moreso asking about checking if a python version exists AND has lark installed in the testing script.
E.g. python3.13 and 3.12 are installed but lark is only on 3.12.

Temperz87:
I've created a merge request to cty:12 main for this

cty12:
> I think Siek was moreso asking about checking if a python version exists AND has lark installed in the testing script.
E.g. python3.13 and 3.12 are installed but lark is only on 3.12, so the testing script should know to use 3.12.

> I've created a merge request to cty:12 main for this

@Temperz87 Thank you! I think it's a good idea to select the highest version of Python with lark installed, as you suggested. I merged your pull request into my fork of Deduce.

jsiek:
Great! Thank you both!

Issue #46, opened 2024-12-26
print 42 bad error


jsiek:
Says undefined variable `suc`, which reveals how numbers are encoded and is not helpful. The fix is to import Nat, but the error doesn't say that.

HalflingHelper:
Should the error do what we're doing elsewhere, where we check if the call is to suc, zero, node, empty, and then have a custom error for those problems?

PR #48, opened 2024-12-31
More helpful error for missing import

HalflingHelper:
Fixing #46


jsiek (approved):
Nice!

In the future we could think about how to generalize this.

PR #51, opened 2024-12-31
add missing params to type_check_term_inst

HalflingHelper:
env, recfun, and subterms were undefined in the method type_check_term_inst.

Here is the file that triggered the error:

```
import List
import Nat
import Option

theorem rev_correct : all T : type, i : Nat, xs : List<T>.
  get(reverse(xs), i) = get(xs, length(xs) - 1 - i)
proof
  arbitrary T : type, i : Nat
  induction List<T>
  case [] {
    suffices ? by definition { length, get }
    ?
  }
  case node(x, xs') suppose IH {
    ?
  }
end
```

Gotta love that python lets you have undefined variables floating around.

### DIFF
--- a/proof_checker.py
+++ b/proof_checker.py
@@ -1760,7 +1760,7 @@ def type_first_letter(typ):
       print('error in type_first_letter: unhandled type ' + repr(typ))
       exit(-1)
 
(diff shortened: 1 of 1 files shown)


jsiek (approved):
Thanks!

PR #53, opened 2025-01-03
update libraries

HalflingHelper:
I just simplified some of the theorems in `Base.pf` and added 2 theorems in `Nat.pf` that helped with working on my `rev_correct` proof.


jsiek:
Thank you Calvin!

PR #56, opened 2025-01-15
Automagically find stdlib!!!!!!

Temperz87:
Make deduce automatically find stdlib
Added --no-stdlib argument just in case you don't want stdlib 
Made tester no longer use --dir as it's not needed

Also, can we get someone to test this on Windows? I used `os.path.join` and all, but I'm very paranoid.

### DIFF
--- a/deduce.py
+++ b/deduce.py
@@ -79,6 +79,10 @@ def deduce_directory(directory, recursive_directories):
 
 if __name__ == "__main__":
     # Check command line arguments
+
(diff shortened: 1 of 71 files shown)


jsiek:
This feature is nice, thank you!

Why has the output changed for errors? There's a line that prints "./lib".
That seems unnecessary. Can you remove that?

Temperz87:
Matei saw this and joking said "did you push a print statement to prod?"
I may have accidentally pushed a print statement to prod, sorry about that. I have since fixed this.

Issue #57, opened 2025-01-16
Document command line arguments for `deduce.py`.


jsiek:
The command line arguments for `deduce.py` are not documented on the Deduce web page.
Please add a section after "Getting Started" and before "Syntax/Grammar Overview" that
explains the command line arguments.

PR #58, opened 2025-01-16
Document command line arguments

Temperz87:
In regards to #57  I've created a rough draft of what the documentation could look like. When writing the docs, I assumed that 2nd year CS students would be the ones reading it, and as such went into a great amount of detail when describing commands. I also accept that some of my explanations are probably not up to par, and will need a second set of eyes on them.


jsiek:
This is a great step in the right direction! Thank you Shulin!

PR #67, opened 2025-01-22
Add markdown doc testing

mateidragony:
I updated the `convert.py` file to basically do exactly what entangled does. It reads the comment at the bottom of the markdown file (that looks like this)

```
<!--
  {.deduce^file=FunctionalProgramming.pf}
  <<ImportNat>>
  <<five_six>>
  <<print_five>>
  
  // extra code I want to add
  import Nat
  
  <<print_five>>
-->
```

and replaces the ids within the `<<` and `>>` signs with the respective labeled code in the markdown file. That gets outputted to a file in the `test/should-pass/` directory prepended with `doc_`.

The code in the `gh-pages/deduce-code` directory is not used for testing. The code there is only used by the site to color the syntax and display in a code block. The only files in that directory that are actually being ran for testing are the `home_example` files which are hardcoded in `test-deduce.py`.


jsiek:
Could you give a high-level description of what this PR does? Like what is the goal? The above comment seems to be talking about implementation details... Does this PR change what happens when someone pushes a commit to the repo?

mateidragony:
This PR generates new test files for all of the deduce code in the doc files and the site every time someone pushes to main.

jsiek:
Does it run deduce on those generated files?

mateidragony:
It adds them to the `test/should-pass` directory so the test-deduce workflow should test them

jsiek:
That's great. Could you break one of the docs on this branch to see if the workflow fails as it should?

mateidragony:
Oh I think I might need to make the pages deploy script rerun the testing script because I think the tests ran before the error-prone test file was generated

mateidragony:
So I updated our test workflow to now trigger after the site has been deployed which should catch any errors in the docs.

jsiek:
Great!

PR #69, opened 2025-01-23
Add threading into the test suite when checking errors

Temperz87:
"--errors" in the context of the test suite [used to take over a minute](https://i.kym-cdn.com/entries/icons/original/000/032/744/maneuver.jpg),  so I added threading and cut that time down to about 30 seconds. I also added some new arguments into the tester:

`--max-threads thread_count` will set the max thread count equal to `thread_count`

`--destroy-my-cpu` will set the max thread count to 999.

I also changed how deduce.py works in two key ways
1. When it catches SIGINT, it'll exit with 137. This allows the tester to see that deduce.py exited with SIGINT and do the same
2. I made it such that if you do "py deduce.py" while in the same directory as deduce, it will work instead of breaking as deduce expects an absolute path as its first argument


Temperz87:
I also would like to point out that the test script was about 40 seconds faster on this PR.

jsiek:
Thanks for this PR!

PR #71, opened 2025-01-23
Testing doc error

mateidragony:
This pr should fail and should not be merged. It is testing that an error in the doc files gets caught by the test workflow.


mateidragony:
yay!

Issue #73, opened 2025-01-24
`list index out of range` error escapes rec_desc_parser


HalflingHelper:
Many programs, such as:

```
define
```

```
theorem blah : true
proof
  have a : true
```

Trigger a python list index out of range error that isn't handled, and gets printed as part of the output

This is only because we're not doing `if end_of_file():` before every single call to `current_token`. I could pretty easily just add an end of file check to the current_token method which should fix the issue, or I could add the checks every place we need to. The former is much easier, but I also don't know whether we want to do that extra check in a method we call 400 times, even though it isn't always necessary.

HalflingHelper:
Fortunately these programs are all trivially and obviously bad, so it shouldn't impact students too much, it's just nice to make sure that python doesn't escape

Issue #54, opened 2025-01-08
precedence for iff vs if-then in lalr vs. recursive descent


jsiek:
This formula parses differently in lalr vs. recursive descent:
```
if (P ⇔ Q) and (Q ⇔ R) then P ⇔ R
```
I'm not sure which should be higher precedence, if-then or iff.

PR #78, opened 2025-01-24
Parser!!!!

HalflingHelper:
This should address issues #73 , #54 

I double checked that the grammar rules in `Deduce.lark` and the structure of the recursive descent parser are the same.

I also added an out of bounds check that errors more gracefully at end of file. (Though eventually it might be nice to provide a little more context than `Expected Token`, maybe an optional argument to `current_token` that gives what is expected, or a different `consume_token()` type method that errors if the token isn't expected.


jsiek:
Excellent, thank you!

Issue #77, opened 2025-01-24
test-deduce crashes because of convert


jsiek:
```
jsiek@sice-mbp-986171 deduce % python3.11 ./test-deduce.py 
Traceback (most recent call last):
  File "/Users/jsiek/deduce/./test-deduce.py", line 7, in <module>
    from doc.convert import convert_dir
  File "/Users/jsiek/deduce/doc/convert.py", line 129
    return f'<a href="{link}#{'' if m.group(3) is None else m.group(3)}" target="{'_self' if is_local_md else '_blank'}">'
                                                                                   ^^^^^
SyntaxError: f-string: expecting '}'
```

jsiek:
I removed the call to `convert_dir` in the default "run everything" mode. Does that break the github action that checks the docs, or does that use the `--site` flag? I hope it does...

mateidragony:
Removing the call to `convert_dir` will make it so that the test script doesn't actually check doc files. The github action doesn't call the test script with any flags so removing `convert_dir` will make it so that the workflow doesn't check doc files. 

I'll create a pr with a fix to the error above in a second.

PR #79, opened 2025-01-25
Fix test bug

mateidragony:
This should address the bug in #77). I also updated the test file to actually run the converted proofs when run with the `--site` flag (which I forgot to add before...)


mateidragony:
Good to go

jsiek:
Thanks!

PR #80, opened 2025-01-27
Add --version argument

Temperz87:
`py deduce.py --version` will now print a funny message that says "Deduce: version 1.0.0"!
I also forgot to commit the `--destroy-my-cpu` argument into the tester, so I'm sneakily adding that as I'm a little embarrased

### DIFF
--- a/deduce.py
+++ b/deduce.py
@@ -82,8 +82,8 @@ def deduce_directory(directory, recursive_directories):
             deduce_directory(directory + file, recursive_directories)
 
 if __name__ == "__main__":
-    # Check command line arguments
     signal(SIGINT, handle_sigint)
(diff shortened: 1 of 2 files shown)


Temperz87:
Also on this note, we might be able to run the GitHub action with max threads, as GitHub's computers going into max load shouldn't be a problem

jsiek:
The name `--destroy-my-cpu` is fun, but not informative. Please rename it to something that helps someone decide whether they want to use the flag, or not.

Temperz87:
Actually it's probably a good idea to just remove this argument and instead pass in `--max-threads 9999` if you want it to be uncapped

jsiek:
Ok, thanks!

PR #81, opened 2025-01-27
Website SEO improvements

mateidragony:
Adding some extra tags to improve search engine optimizations for the site. This will hopefully make it easier for search engines to index the site (DuckDuckGo already has!) and make the result look better.

This is what the first search result is on duck duck go (for "jsiek deduce"):

image

The title `deduce | A proof checker meant for education. Primarily for teachi...` is taken from the name of the repo and description on github btw.


jsiek:
Thank you!

Issue #76, opened 2025-01-24
Command line argument to suppress `.thm` files


mateidragony:
Sometimes I don't really want a theorem file...

mateidragony:
If we do this (and for all new command line args to deduce) we should remember to update `doc/GettingStarted.md` to document those args

mateidragony:
We might also want deduce to check if there's even anything to put in a `.thm` file before it creates one. I've got a couple of empty `.thm` files that got generated for files with just asserts or functional code

PR #86, opened 2025-01-29
Make theorem files conditionally generate

Temperz87:
In reference to #76 , I've made sure that theorems only generate if the file wouldn't be empty. I've also added the `--suppress-theorems` argument, which turns off all theorem generation when passed


Temperz87:
I think I messed up the documentation

mateidragony:
What makes you say that?

mateidragony:
It's fine

mateidragony (commented):
Looks good!

PR #89, opened 2025-01-29
Private keyword!!!!

### LINKED ISSUE #4: public/private with respect to import
Provide a way to declare something as private to the current file so it can't be imported.

Temperz87:
You can now do `private union` or `private define` or `private function` and it won't get exported (and we can finally close #4). One downside though, is I have no idea how to write tests for this without making my own directory in tests called "private-tests", although I think that I'm just going to do that.

Also, this isn't ready to merge yet as tests are nonexistent


Temperz87:
I also need to update the docs to document this

jsiek:
Why do you need a new directory for testing `private`? (I'm hoping that is not necessary.)

Temperz87:
Because I need a way to import a file that correct parses, which would have to be in the same directory as should-error. However, everything in the should-error directory is currently assumed to have an error, so I'd have to somehow make an exception for that particular files (or get rid of the "no .pf.err" message). I could alternatively do some funky "--dir" stuff for imports, but that'd require creating an exception once more

jsiek:
How about adding `private` to `find_quotient` in `lib/Nat.pf` and then testing to make sure you can't refer to `find_quotient`?

jsiek:
Another one is `L1` in `lib/ListTests.pf`.

Temperz87:
NatTests.pf uses `find_quotient` so i can't make it private, however making stuff private in the NatTests and ListTests files should work

Temperz87:
The error message for using something that's private right now is "Undefined variable" instead of "Trying to use private variable", does trying to use a Lemma have a better message?

Temperz87:
This pr should be ready now

test-deduce.py
jsiek, line 1: Are these changes related to the PR, or are they orthogonal?
What are these changes accomplishing?

Temperz87: Actually I should revert these changes and create a new pr explaining this, I just got a bit carried away when I looked at the tester code

jsiek: It's easy to get carried away. Thanks for reverting.

doc/index.en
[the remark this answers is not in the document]
Temperz87: I don't know where this came from, so I've removed it

jsiek:
This PR is excellent, thank you Shulin!

PR #91, opened 2025-01-29
Test more stuff at the same time

Temperz87:
Due to the nature of how imported files get processed, it's better to call Deduce with more files than less files, which is what this commit makes the tester do. Previously, when testing everything, Deduce was being called six times. However, Deduce gets called twice when testing everything (once for errors and once for testing passables), which speeds up the tester a lot.

### DIFF
--- a/test-deduce.py
+++ b/test-deduce.py
@@ -228,7 +228,7 @@ if __name__ == "__main__":
     if test_errors:
         test_deduce_errors(deduce_call, error_dir)
 
-    if not (test_lib or test_passable or test_errors or test_site) == 1: # run everything
+    if not (test_lib or test_passable or test_errors or test_site): # run everything
         # test
         test_deduce(parsers, deduce_call, [lib_dir, pass_dir, site_dir + '/home_example1.pf',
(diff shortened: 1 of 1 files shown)


test-deduce.py
```diff
@@ -225,17 +228,11 @@ def test_deduce_errors(deduce_call, path):
...
 
-    if len(sys.argv) == 1: # run everything
-        # test the home examples
-        test_deduce(parsers, deduce_call, site_dir + '/home_example1.pf')
-        test_deduce(parsers, deduce_call, site_dir + '/home_example2.pf')
-        test_deduce(parsers, deduce_call, site_dir + '/home_example3.pf')
-        # generate test files for doc code without generating html
-        # convert_dir("./doc/", False) # Requires markdown to be installed
-        
+    if not (test_lib or test_passable or test_errors or test_site) == 1: # run everything
```
jsiek, line 231: Why `1` and not `True`?

Temperz87: I forgot to remove that, however for some reason in python `True == 1 => True` , which is weird to me

PR #93, opened 2025-01-30
Refactoring all parser errors to be `ParseError`

HalflingHelper:
Does what it says on the tin! Making everything consistent meant that some errors that weren't done consistently in the old paradigm changed a little bit, but nothing major wrt the content of the error.

This also let me remove a good number of the error helper functions (last, missing, extend), and everything should be handled by the subclass.

### DIFF
--- a/error.py
+++ b/error.py
@@ -83,6 +83,7 @@ class StaticError(Exception):
 def static_error(location, msg):
   raise StaticError(error_header(location) + msg)
 
+MAX_ERR_DEPTH = 2
 
 # Parse Errors need to carry around some extra data
 class ParseError(Exception):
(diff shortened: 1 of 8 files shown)


HalflingHelper:
I'll also note that as I briefly mentioned last meeting that I started writing tests to catch every parsing error. Though those are only halfway done and not ready to merge yet, this branch passed perfectly all of these, which is encouraging in terms of making sure that this change doesn't change behavior.

HalflingHelper:
Also worth putting that the reason for this refactor is twofold, first because I think it is good for the parser to treat error messages consistently, and additionally to make it easier for the language server to be able to interact with parse errors uniformly.

PR #98, opened 2025-02-03
Fixes #97 and handles empty files

### LINKED ISSUE #97: Broken Block Comments
```
/* Here is a Comment
********/

 print true

 /* Here is a Comment
********/
```

Both parsers read this as one big comment, causing lalr to error, and recursive descent to not print anything

HalflingHelper:
Yay!


jsiek:
Excellent, thank you!

PR #100, opened 2025-02-04
Add theorems into the release

Temperz87:
In the package target in the Makefile, add the line `$(PYTHON) ./deduce.py ./lib` which generates the `.thm` files into the folder.


jsiek:
Thanks Shulin!

PR #102, opened 2025-02-04
Update markdown converter to fix block bug

mateidragony:
This fixes the bug with the markdown converter where new code blocks and lists sometimes weren't detected.

For example

```
Some markdown text
'''
code block
'''
```

Wouldn't render the code block and the generated html would look wrong. Instead of requiring us to add an explicit new line between text and code blocks, the converter detects the block and renders it accordingly (as most markdown renderes do).


jsiek:
Thanks Matei!

Issue #95, opened 2025-01-31
Add ascii equivalent for suffices wildcard `─`


mateidragony:
It would be nice to have an ascii equivalent for `─` in suffices proofs maybe an underscore or something.

jsiek:
Note to self: there's a bug regarding the omitted term and suffices... it showed up in the proof of `int_mult_assoc`. Change a suffices to use  ─ and you get a funky error message that talks about `if`-`then`.

PR #119, opened 2025-02-07
New suffices omitted term `__`

mateidragony:
This adds `__` as an alternative to `─` for omitted terms in suffices statements #95.

Since `__` is now lexed as an `omitted_term` we can no longer use `__` as an identifier (hence the new error file I added). We can also workshop a different ascii alternative so we don't have that issue.


jsiek:
Thanks!

Could you also add an entry in `Reference.md` for this, including an example?

mateidragony:
This is what I wrote in the reference file (open to change):

## Suffices Omitted Term

Use suffices without explicitly stating the new goal. Instead, you can use `─` or `__` to fill in what the new goal should be. 

Example:

```{.deduce^#suffices_omitted_example}
theorem length__nat_one': all x:Nat. length([x]) = 1
proof
  arbitrary x:Nat
  suffices __
      by definition {length, length}
  add_zero[1]
end
```

This acts similarly to [`?`](#question-mark--proof), except that deduce will fill in the omitted term internally and accept the proof as complete.

jsiek:
Thanks!

PR #122, opened 2025-02-09
Tweaks

HalflingHelper:
- Fixed position tracking for RecFun statements to enclose the whole definition
- Replaced `begin` with `proof` in the `__str__` method for theorems
- Added a space after arbitrary when printing

### DIFF
--- a/abstract_syntax.py
+++ b/abstract_syntax.py
@@ -1874,7 +1874,7 @@ class AllIntro(Proof):
     x, t = self.var
     res = ''
     if s + 1 == e:
-      res += 'arbitrary'
+      res += 'arbitrary '
     res += f"{x} : {str(t)}"
     if s == 0:
       res += ";"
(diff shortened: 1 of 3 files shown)


HalflingHelper:
Here's where arbitrary was pirnting wrong:
```
theorem suc_one_add: all n:Nat.
  suc(n) = 1 + n
proof
  arbitrary n:Nat
end
```

```
/home/calvin/Documents/Programming/343/deduce/lib/Nat.pf:193.1-193.4: missing conclusion after
        arbitraryn : Nat;None
/home/calvin/Documents/Programming/343/deduce/lib/Nat.pf:186.1-189.18: while parsing
        proof_stmt ::= "theorem" identifier ":" formula "proof" proof "end"
```

HalflingHelper:
Not that I mention it, the None is a bit odd too? Should we be doing null checks in every `__str__` method for nonterminals? Just in case?

jsiek:
thats a good idea

jsiek:
thanks fir these fixes!

Issue #127, opened 2025-02-11
Automatically add assumptions to env for induction and switch cases


HalflingHelper:
If there is no `assume IH` or `assume prop` after `case`, add the thing that would be assumed anyway as an unnamed premise.

PR #134, opened 2025-02-11
Automatic induction and switch assumptions

mateidragony:
Addresses #127 

For induction deduce now adds in however many un-assumed assumptions (as underscores) needed to match the number that there should be.

Since switch can only have one assumption, if there is none deduce will add one (as an underscore)


jsiek:
Excellent!

PR #142, opened 2025-02-12
Eq sym proof

mateidragony:
Proof of the symmetry of equality. 

I was working on BinarySearch and needed to show that `(first(n) = x) = false` = `(x = first(n)) = false`

### DIFF
--- a/lib/Base.pf
+++ b/lib/Base.pf
@@ -120,15 +120,15 @@ theorem eq_sym : all T:type. all x:T, y:T.
 proof
 	arbitrary T:type, x:T, y:T
 	switch (x = y) {
-		case true {
-			have: x = y by apply eq_true to recall (x = y) = true
+		case true assume xyt {
+			have: x = y by apply eq_true to xyt
 			replace recall x = y
 		}
-		case false {
-			have: not (x = y) by apply eq_false to recall (x = y) = false
+		case false assume xyf {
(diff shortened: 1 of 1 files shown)


jsiek:
I think this is what `symmetric` is for... so unless I'm missing something, we don't need that theorem.

mateidragony:
How should I use symmetric to get from `(first(n) = x) = false` to `(x = first(n)) = false`?

```
have: (x = first(n)) = false by {
	have: (first(n) = x) = false
	rewrite eq_sym<Nat>[first(n), x] in recall (first(n) = x) = false
}
```

This was my have with the only relevant given being `(first(n) = x) = false`

jsiek:
You can use `eq_false` from `Base.pf`. Perhaps also `iff_eq`.

mateidragony:
I got it, I just needed to do this to get `x=first(n)` to be able to start doing some rewrites:

```
suffices not (x = first(n))
	by rewrite symmetric (apply iff_equal to eq_false[x = first(n)])
assume G
```

It felt a bit annoying, but that difficulty came mostly from the order in which I was trying to prove things. If my have was instead `not (x = first(n))` I would've saved myself a lot of trouble.

We could just chalk this all up to user error and discard of that theorem...

jsiek:
Is there some documentation that we could add that would have helped steer you in the right direction?

Issue #124, opened 2025-02-10
Bad error messages when uniquified names collide


HalflingHelper:
Deduce isn't clear about the actual issue when there are multiple variables with the same name floating around in proofs.

I think we could either warn when doing a case with an already used name, or just add a check when we throw this particular error,

```
import Nat
import List

// problem 4
theorem length_take: all T:type. all n:Nat, xs:List<T>.
  if n ≤ length(xs)
  then length(take(n, xs)) = n
proof
  arbitrary T:type
  induction Nat
  case 0 {
    evaluate
  }
  case suc(n') assume IH {
    arbitrary xs: List <T>
    switch xs {
        case [] {
          evaluate
        }
        case node(n', ns') {
            assume prem : suc(n') ≤ length(node(n', ns'))
            ?
        }
    }
  }
end
```

```
/home/calvin/Documents/Programming/343/deduce_labs/test.pf:21.31-21.33: expected a term of type Nat
but got term n' of type T
```

jsiek:
I like the idea of adding warnings for shadowed variables.

HalflingHelper:
How do we feel about messages like this (in uniquify)
```
/home/calvin/Documents/Programming/343/deduce/lib/asdf.pf:4.25-4.39: WARNING: T already in use
/home/calvin/Documents/Programming/343/deduce/lib/asdf.pf:9.11-9.12: expected a term of type T
```

HalflingHelper:
How do we feel about ANSI colors?

![Image](https://github.com/user-attachments/assets/8287bafe-d5b7-4926-9fee-9c05be940808)

HalflingHelper:
Also there are a lot of places where these overwrites happen, but warning every single place might be a bit much.

Like for example I don't feel like it's necessary in an all formula?

mateidragony:
Ansi colors could be an issue itself because we can really spruce up error messages by coloring certain things that should standout

mateidragony:
I do like the idea of warnings telling students that what they're doing technically works but is disadvisavle

HalflingHelper:
Looking at where it happens in Nat  (ignoring '_' instances), it looks to either be from having `some m` when `m` is in the goal and environment, or from renaming local haves the same as already used things (namely `equal` and `sqm_le_n`)

There's one more point where it happens because of shadowing in function cases.
```
recursive dist(Nat, Nat) -> Nat {
  dist(0, n) = n
  dist(suc(m), n) =
    switch n {
      case 0 {
[...]

PR #145, opened 2025-02-15
Warnings for shadowing in uniquify

HalflingHelper:
Adds warnings in uniquify when you shadow a variable, as well as catching attempting to overload theorems and unions post-facto


HalflingHelper:
#124

jsiek:
Excellent!

Issue #159, opened 2025-02-20
Weird `/lib` files


mateidragony:
We've got some files in the standard lib that don't look like they belong there.

- There's a couple test files (`ImportTests.pf`, `IntTests.pf`, `ListsTest.pf`, `NatTests.pf`)
- Then we have `sums.pf` which looks like it just has a couple of example theorems

Do we want to move the test files to the `/test` directory and `sums.pf` to `/ex` or should we leave them in `/lib`

PR #162, opened 2025-02-20
📚 Update lib

mateidragony:
Addresses #159 

- Renamed `sums.pf` to `Sums.pf`
- Moved lib test files to `/test/should-validate`


mateidragony:
So I now understand why those test files were in `lib`. It was to test that private imports fail. So instead I'm importing from other lib files.

mateidragony:
So I put the private bit union with a private function in `Binary.pf` and private variables in `Log.pf`. This works but now we have an unused function and union in `Binary.pf` just hanging around. A better fix for this would be to update the tester to allow importing non lib files to test that private fields can't be imported, or we could start using private functions, variables, and unions for real in the library, then write tests to check that we can't import those.

jsiek:
Thank you Matei!

Issue #165, opened 2025-02-22
Python Error: `AttributeError: 'PatternBool' object has no attribute 'parameters'`


HalflingHelper:
```
theorem idek : all A : bool, B : bool, C : bool.
  if (switch A { case true { false } case false {false}}) then C

proof
  arbitrary A:bool, B:bool, C:bool
  suppose prem
  have asdf : A
    by prem
  ?
end
```

I'm on it, just need to match on the pattern types

HalflingHelper:
Fixed by #166

PR #166, opened 2025-02-22
#165

HalflingHelper:
EZ


HalflingHelper:
The some copy thing was just another weird parsing error I came across in programs like this
(ignore the actual proof, the point is that this should not trigger a python error during parsing :)
```

theorem build_gcd : all a : Nat, b : Nat, d : Nat.
  if is_gcd(d, a, b) then
  (some x : Nat, y : Nat. b * x - a * y = d)
proof
  induction Nat
  case 0 {
    arbitrary b : Nat, d : Nat
    suffices (some x : Nat. d * x = 0) by evaluate
    ?
  }
  case suc(a') assume IH {
    arbitrary b : Nat, d : Nat
    // suppose prem : is_gcd(d, a, b)
    // have prem' : ? by definition is_gcd in prem
    // have lt : (all d':Nat. (if (divisible(d', a) and divisible(d', b)) then divisible(d', d)))
      // by definition is_gcd in prem
  }

end
```

Issue #139, opened 2025-02-12
forgeting `rewrite`


jsiek:
Students often forget to put `rewrite`. Perhaps we can improve the error message to remind them to use it. Maybe handle this in `check_implies`?

HalflingHelper:
The error is also pretty badd when forgetting definition

It asks 'did you mean recall', but I think we have all the information to look up what the thing is in the environment, and intelligently determine advising definition vs rewrite as the advice

HalflingHelper:
The problem with doing this in `check_implies` is that the most we could say is `Did you mean replace?`
That function has no knowledge of what the user actually typed

mateidragony:
This may be controversial, but what are our thoughts on disallowing writing just a proof variable. Like we force them to use 'replace' and add a new keyword for applying proofs straight up. (I may not have explained this well)

mateidragony:
Like to use theorem P you can't just write P in your proof you need to say 'using P' or something like that

HalflingHelper:
It's an interesting idea, personally I don't know how helpful it would be, because people would probably still forget to use `replace` or whatever (it's not like we really taught `PVar` this semester).

If we did `using` or whatever, it then becomes a syntax error which I guess could be easier for students to decipher?

mateidragony:
It would make error messages better because we can always catch and say if they forgot using or replace

HalflingHelper:
Regardless I think that we still want the code addressing this issue, your idea would be separate IMO.

HalflingHelper:
I don't think that more keywords is always a solution

PR #167, opened 2025-02-23
Forget def / recall

HalflingHelper:
Advice for when 'replace' or 'definition' could be used (#139)

As opposed to the suggestion of the issue, I do this in the final case of `check_proof_of`, which handles (among other things) AllElim, PVar, and Recall nodes, and offer the advice when the proof fails, and the attempted proof was equational in nature.

The benefit of doing the check here, rather than in `check_implies` is that there is more information available about the actual form that the used typed, rather than check_implies, which has only `form_red`, and check implies also handles many more cases so the behavior is harder to control and predict there.

I also updated `Env._formula_of_proof_var` to  check if the user is mistakenly trying to use a function-typed Term Var as a proof var, and suggest definition, rather than recall in that case.

### DIFF
--- a/abstract_syntax.py
+++ b/abstract_syntax.py
@@ -1,7 +1,7 @@
 from dataclasses import dataclass, field
 from lark.tree import Meta
 from typing import Any, Tuple, List
-from error import error, warning, set_verbose, get_verbose, get_unique_names, VerboseLevel, EnvLookupException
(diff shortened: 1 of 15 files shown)


HalflingHelper:
`EnvLookupException` might now be unnecessary actually, and a holdover from earlier attempts to do this

jsiek:
Excellent!

Issue #155, opened 2025-02-17
Adga-style stdlib web pages


mateidragony:
It would be nice to have the Deduce standard library files viewable from the website, similar to how agda does it ([example: Agda.Builtin.Nat](https://agda.github.io/agda-stdlib/master/Agda.Builtin.Nat.html)).

We could set it up to have the pages be autogenerated from the `/lib` files with syntax highlighting and links inserted to functions defined in different parts of the stdlib.

PR #168, opened 2025-02-23
🌍 Site stdlib + more

mateidragony:
The main point of this pr is to address #155. Here is a preview of what it will look like on the site https://mateidragony.github.io/deduce/pages/stdlib.html

I also found a way to not have to push generated html files to the repo while still deploying them to GitHub Pages. This is really nice as I'm generating a lot of html for the stdlib and #156 will require a lot of additional html that could've cluttered our repo. All of the generated html files were found in `/gh-pages/pages` so that directory can be deleted from the repo.
 
One quick point on the stdlib file generation:
- I made the generation script use lark to tokenize the library files for syntax coloring. Since I'm expecting future parser changes, I made it so that the generation script has a list of `know_tokens` and queries our current `Deduce.lark` file to see if new tokens have been added that it doesn't know about. If so, the script errors and prompts us to update the `know_tokens` list accordingly. (`known_tokens` maps from lark token type to syntax coloring type (prim, keyword, etc.))


jsiek:
Wow, that was fast!

Issue #116, opened 2025-02-07
opaque definitions


jsiek:
An `opaque` definition is public but outside of its defining module, you're not allowed to apply the definition using `definition`, which means the internals of the definition are hidden.

PR #178, opened 2025-03-06
Accessibility refactor and opaque stuff!!!!!!!!!!!!!!!!!!!!!!!!!!

Temperz87:
This PR adds the ability to create an opaque "declaration", and also refactors how accessibility gets parsed.

The modified syntax is:
```
access_decl ::= "private" declaration | "opaque" declaration
declaration ::= union | rec_function | function | definition

statement ::= access_decl | declaration | ...
```

If something is declared as "opaque", then you cannot do "definition ..." with it inside of a proof. 

Also, a lot of error messages have changed because "statement ::=" has become "declaration ::=" I skimmed through to make sure I didn't change anything weird, however I could have missed something.


Temperz87:
Also with this PR we can close #116

mateidragony:
Chart parser got pushed?

mateidragony:
Also `actual_error.tmp`

actual_error.tmp
mateidragony, line 1: Delete, and we should prolly have something in `.gitignore` to not push `*.tmp` files

chart-parser/grammar_interpreter.py
mateidragony, line 1: spoliers!

Temperz87: 'git add .' and its consequences should be studied

jsiek:
I'll take a look soon.

jsiek:
@Temperz87 I was in the middle of some changes... sorry, you'll need to merge.

proof_checker.py
```diff
@@ -2462,7 +2462,11 @@ def process_declaration(stmt, env, module_chain):
   if get_verbose():
     print('process_declaration(' + str(stmt) + ')')
   match stmt:
-    case Define(loc, name, ty, body, isPrivate):
+    case Define(loc, name, ty, body):
+      if stmt.is_opaque():
+          error(loc, 'definition ' + base_name(name) + ' cannot be used inside of a proof')
```
jsiek, line 2467: I don't understand why this `if` and error message is here. What's an example Deduce program that would trigger it?

test/should-error/opaque_use.pf.err
```diff
@@ -0,0 +1 @@
+./test/should-error/opaque_use.pf:1.1-1.19: could not find a file for import: ImportTests
```
jsiek, line 1: This looks like the wrong error message... it's not actually testing what it is suppose to.

jsiek (commented):
Shulin, great progress! Please take a look at my comments are respond.

jsiek:
From the description of this PR:
```
If something is declared as "opaque", then you cannot do "definition ..." with it inside of a proof.
```

That's kinda wrong... if something is declared as "opaque", then you cannot do "definition ..." with it in a file that is different from the one that the thing was declared in.

test/should-error/opaque_use.pf.err
Temperz87: I missed up here, this file needs to be remade as `ImportTests` has been removed

proof_checker.py
Temperz87: `definition thing` where `thing` is opaque should trigger it. The test case which showcased this got altered while I was merging, so I'll create another

Temperz87:
If I have an opaque union 
```
union Nat {
    zero
    suc(Nat)
}
```

What should the intended behavior be? I.e. should I see `suc` inside of a proof? If I shouldn't, should I see `zero` inside a proof? My thought with opaque was that you shouldn't be able to do definition with them, which is the current behavior, however with hearing that `Nat` hopefully will be opaque one day, I'd like to understand how opaque should be interacting with unions.

proof_checker.py
jsiek: That's the wrong AST node for `definition thing`. The right ones are `ApplyDefs`, `ApplyDefsGoal`, `ApplyDefsFact`.

Temperz87:
If set becomes opaque, how would one do the `empty_union` proof outside of `Set.pf`, or is the point that you shouldn't need to if anything that's opaque comes with the correct stuff?

jsiek:
The opaque stuff has to come with all the right theorems.

Temperz87:
@jsiek Provided that the tester doesn't fail, this should be ready to go.

The only thing I'm not confident about is my ability to write good error messages. I may have made some of them either too techincal or too wordy.

Temperz87:
Oh I see a problem now.

Doing `assert opaque_thing = somethingelse` isn't going to work as we don't know if the variable is in the proof context or not.

Temperz87:
Alright so I just made a global variable that's "error if we try to reduce an opaque", so we should be fine now.

Temperz87:
Okay so basically I just found a huge problem

```
theorem isomorphic: all x:Nat_Hat. natToNatHat(NatHatToNat(x)) = x
proof
    ?
end
```

Gives the advice:
```
        Alternatively, you can try induction with:
                induction Nat_Hat
                case cero {
                  ?
                }
                case next(N1) assume IH1: natToNatHat(NatHatToNat(N1)) = N1 {
                  ?
               }
```          
Which doesn't work as `Nat_Hat` is opaque

Temperz87:
Alright I _think_ I've covered everything, however I'd like to sit on this for a day because I'm sure some issue will come to me in a dream.

Temperz87:
I can't think of any more issues, so I think it's good @jsiek

PR #179, opened 2025-03-10
Create `CONTRUBUTING.md` and other `docs`

mateidragony:
Created the `/docs` directory to store any Deduce contribution documentation. So far we have documentation about:
- Our code of conduct (adapted from the [Contributor Covenant homepage](https://www.contributor-covenant.org/version/2/1/code_of_conduct.html))
- Bug reporting
- Suggesting enhancements
- Contributing code

> Note: Since the `/doc` directory was moved to `/gh-pages` and the test script imports `convert.py` from the `/doc` directory, I had to rename `/gh-pages` to `/gh_pages` since python does not let you import files from directories with hyphens....

I also created a `knowlegde-base` directory within the docs where we should put all of the information on how Deduce works internally, and how to maintain Deduce. What is in `knowledge-base` so far:
- `how-it-works`: This is empty for now (I will create an issue to fill this in), but this file will contain information on how deduce works. I kinda stole this idea from the docs for [Angband](https://github.com/angband/angband/blob/master/docs/hacking/how-it-works.rst)
- `what-to-update.md`: This file contains information on what internal (and external) parts of Deduce need to be updated if you're changing some part of Deduce.
- `site-maintenance.md`: This file was originally in `/gh_pages` but I moved it here to keep documentation all in one place.


mateidragony:
A link to my fork for easy file viewing https://github.com/mateidragony/deduce/tree/doc-migration/docs

jsiek:
Thanks Matei!

Issue #180, opened 2025-03-12
Missing conclusion doesn't work inside of cases, maybe elsewhere


HalflingHelper:
```

recfun fib(n : Nat) -> Nat
  measure n
{
  if n < 2 then n
  else fib(n - 1) + fib(n - 2)
}
terminates {
  arbitrary n:Nat
  have A: if not (n < 2) then n - 1 < n by {
    assume prem: not (n < 2)
    switch n {
      case 0 assume eq_z_t {
        ?
      }
      case suc(n') assume eq_sn_t {
        ?
      }
    }
  }
  ?
}
```

This gives this error
```
/home/calvin/Documents/Programming/343/deduce/test/should-validate/fib.pf:18.7-18.11: expected a proof.
Did you mean "cases" instead of "case"?
/home/calvin/Documents/Programming/343/deduce/test/should-validate/fib.pf:4.1-17.8: while parsing
        statement ::= "recfun" identifier type_params_opt "(" variable_list ")" "->" type "measure" term "{" term "}" "terminates" proof
```

I think there are just more places where holes can be that we're not checking

Issue #158, opened 2025-02-19
error message position mixup for incomplete proof (missing conclusion)


jsiek:
```
theorem T: all P:bool, Q:bool. if P and (if P then Q) then Q
proof
  arbitrary P:bool, Q:bool
  assume prem: (P and (if P then Q))
end
```

error:
```
/Users/jsiek/deduce/test/should-error/suffices_implies_hole.pf:3.23-4.37: incomplete proof
Goal:
	Q


Givens:
	prem: (P and (if P then Q))
```
The `3.23` points to the second `bool` in the `arbitrary` statement, which is rather odd.

HalflingHelper:
Yeah, that was my way of trying to highlight the last statement in the proof

I can change it to highlighting the `end` token

PR #182, opened 2025-03-13
Fixing missing conclusion issues, test-deduce bug, and adding directory for test imports

HalflingHelper:
### Conclusion issues
Addresses #158 and #180 related to the parser. Specifically changing the location given to be the end keyword of the proof, or whatever else the next token is, and adding the automatic PHole in empty proofs, which wasn't being done previously -- a test case was added for this

### Test-deduce script
We were removing from the list of threads as we iterated over it, leading to not every error message getting tested. 

### test-imports directory
Added a directory under `/test/` for files that tests for private and opaque can import. These files aren't run themselves, but `test-deduce.py` adds it as a `--dir` argument, so those files are accessible to files in the `should-error` directory


jsiek:
Thanks Calvin!

Issue #181, opened 2025-03-12
Remove enable


Temperz87:
Enable is deprecated and evaluate should be used in its place.

PR #184, opened 2025-03-15
emable is gone 🥳

HalflingHelper:
#181 

Someone might want to double check the example for working with not (the proof of less_irreflexive) since it now relies on less_suc_iff_suc_less, which makes things not quite so clear. I didn't see any quick examples in the docs, but maybe we can find some other example.

(Maybe proving `if x then not (not x)` would work)


jsiek:
excellent, thank you Calvin!

PR #190, opened 2025-03-21
Exponents

HalflingHelper:
Added exponents to Nat.pf, and some intro theorems.

Also fixed parsing to be left-associative for arithmetic (This matters for `-` and `/`)


jsiek:
Excellent, thank you Calvin!

Issue #175, opened 2025-03-05
bad error for `definition eq` where `eq` is a proof variable (and not a term variable)


jsiek:
```
expected a term or type variable, not dec_zero
```

HalflingHelper:
Do you have an example that triggers this? I'm not able to recreate it

jsiek:
Thank you for looking into this!

Yes, so here's a minimal example:
```
lemma L1: true
proof
  .
end

lemma L2: true
proof
  definition L1
end
```
and the current error message is:
```
/Users/jsiek/def_proof.pf:8.3-8.16: expected a term or type variable, not L1
```

I think part of the problem has been fixed, which was the missing position information in the error.
But perhaps there is still room for improvement in that the error message so that it can better teach the
user about how to use `definition` versus `replace`. Oh, I remember now. The part about `type variable` is not so great... this error messaage is a generic one coming from inside an `Env` method, but it would be better to have an error message the gets generated using the knowledge that it is a `definition` proof statement that was in error.

HalflingHelper:
Related is the error message here
```
lemma L1: true
proof
  .
end

lemma L2: true
proof
  definition empty
end
```

```
/home/calvin/Documents/Programming/343/deduce/blah.pf:10.3-10.19: Cannot infer type arguments for
        empty
Please make them explicit.
```

Which asks for explicit type args - which shouldn't be possible in a define. (And the real error is trying to do definition empty) Both that error and the other error are coming from `type_synth_term`

We might need to give additional context to type synth term to get better error messages

HalflingHelper:
I was able to remove the call to `type_synth_term` before doing apply_definitions, and the tests all still passed, do you have an example where the type_synth is needed? I tried messing with overloads and couldn't find any

HalflingHelper:
I guess I should also specify that we still need to do a check to make sure the thing exists in the environment or error messages get bad, but on correct deduce programs it isn't causing issues.

PR #191, opened 2025-03-21
More informative errors in definition

HalflingHelper:
I refactored some code into apply_definitions, and fixed some miscellaneous error message bugs, in addition to addressing #175 with this message
```
/home/calvin/Documents/Programming/343/deduce/blah.pf:8.3-8.16: Expected a term or a type variable when attempting to use the definition of L1.
        If L1 is a theorem or a lemma, you might want to use 'replace'
```

### DIFF
--- a/abstract_syntax.py
+++ b/abstract_syntax.py
@@ -3603,6 +3603,13 @@ class Env:
     else:
       return None
 
+  def _term_var_defined(self, curr, name):
+    if name in curr.keys():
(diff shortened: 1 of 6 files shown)


HalflingHelper:
Seeing an issue actually... hold off on merging

HalflingHelper:
Should be good!

jsiek:
Thank you Calvin, this looks good!

Issue #189, opened 2025-03-20
Error in check_type


HalflingHelper:
Truly horrifying error messages are being seen

Here is my code:
```
theorem summation_dist :
  all k : Nat, a : Nat, f : fn Nat->Nat, s : Nat.
  a * summation(k, s, f) = summation(k, s, fun x { a * f(x)})
proof
  induction Nat
  case 0 {
    // arbitrary a : Nat, f: fn Nat->Nat, s: Nat
    suffices ? by evaluate
    rewrite mult_zero
  }
  case suc(n') assume IH {
    ?
  }
end
```
And here is the output, `--traceback` isn't doing anything for me

```
error in check_type: unhandled type Union(location=<lark.tree.Meta object at 0x7e946bd553d0>, name='Nat.95', type_params=[], alternatives=[Constructor(location=<lark.tree.Meta object at 0x7e946c311fd0>, name='zero.96', parameters=[]), Constructor(location=<lark.tree.Meta object at 0x7e946c09bfe0>, name='suc.97', parameters=[Var(location=<lark.tree.Meta object at 0x7e946c310170>, typeof=None, name='Nat.95', resolved_names=['Nat.95'])])], isPrivate=False) <class 'abstract_syntax.Union'>
```

HalflingHelper:
Changing the suffices to this makes it work
```
 suffices all a : Nat, f : fn Nat->Nat, s : Nat. a * 0 = 0 by evaluate
```

Is the function type the issue?

HalflingHelper:
Here is a more minimal example - 
```
union A { a b }

fun blah(n : A) {
  if n = a then b else a
}

theorem asdf : all x : A.
  blah(blah(x)) = x
proof
  suffices ? by 
  evaluate
  // definition blah
end
```

The problem is that when in reduce_all mode for evaluate, the `Var` node for `A` reduces to the `Union` node for `A`, like this
`type_check_term: (all x.5:union A.0<> {a.1 b.2}. switch switch x { case a { b } case b { a } } { case a { b } case b { a } } = x) : bool?`, versus definition which yields this
`(all x.5:A. switch switch x { case a { b } case b { a } } { case a { b } case b { a } } = x) : bool?`

I was able to fix this by preventing `Var` from reducing when it is bound to a union in the environment, I don't know if there would be a more resilient fix, or whether there would be other times we don't want a var to reduce, even in reduce_all mode

HalflingHelper:
In class Var in abstract_syntax.py:
[...]

PR #192, opened 2025-03-23
Fixing #189

HalflingHelper:
See #189

### DIFF
--- a/abstract_syntax.py
+++ b/abstract_syntax.py
@@ -750,6 +750,8 @@ class Var(Term):
         if res:
           if get_verbose():
             print('\t var ' + self.name + ' ===> ' + str(res))
+          if isinstance(res, Union): 
+            return self
           return res.reduce(env)
         else:
           return self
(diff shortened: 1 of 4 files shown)


jsiek:
Another possibility is adding a case for Union in check_type... I'm not sure which approach is better. There are other AST nodes such as RecFun that we go the other way with.

HalflingHelper:
I thought about that too, but didn't see see how to do it as concisely. Is it just an env lookup that I don't know how to do? And wouldn't that be just turning the union back into some var anyway?

jsiek:
I think the case for `Union` in `check_type` could just `pass` and not do anything because the union would have already been checked at its point of definition.

HalflingHelper:
That by itself won't work, you would need to rewrite a lot of the checking I think to allow the union AST node to work where currently we're using whatever else to represent the type
```
/home/calvin/Documents/Programming/343/deduce/blah.pf:4.10-4.11: expected a term of type union A.0<> {a.1 b.2}
but got term a of type A
```

HalflingHelper:
Is it a TypeInst with no params?

jsiek:
No, I would guess that the `A` in the above error is a `Var` node. But you're correct in that we'd need to add more code, such as more cases in type equality checks, which is unpleasant.

Thanks for answering these questions. I'm OK with this PR now.

PR #196, opened 2025-03-24
❌ Remove `deduce-code` directory

mateidragony:
The `deduce-code` directory doesn't need to be on github since it is generated code.

Making this a pr to see if tests still pass before auto pushing to main...


mateidragony:
Oop fixing now...

mateidragony:
So for now we still need the `home_example` code blocks in deduce code, but once we get index generation we'll be able to completely remove the directory.

PR #203, opened 2025-03-31
Add a theorem into Nat

Temperz87:
This PR adds a theorem saying `pred(suc(n)) = n`

Yay!

### DIFF
--- a/lib/Nat.pf
+++ b/lib/Nat.pf
@@ -373,6 +373,17 @@ proof
  definition pred
 end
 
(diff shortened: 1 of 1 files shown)


Temperz87:
Actually I found out that you can just do 

```
theorem pred_suc_n: all n:Nat. pred(suc(n)) = n
proof
  arbitrary n:Nat
  evaluate
end
```
I feel like at this point the proof should be omitted, however if not I've commited this.

jsiek:
This theorem is still sometimes useful... sometimes you only want to `replace` with this equation and not evaluate everything else.

PR #212, opened 2025-04-22
Evaluation mode using environment instead of substitution

HalflingHelper:


jsiek:
Thank you Calvin! This looks good.

Issue #208, opened 2025-04-16
More testing coverage of parser errors


HalflingHelper:
There are a lot of parse errors that aren't checked by the current set of tests. Create a set of files targeting each specific parse error message.

jsiek:
Working on this — picking a tractable slice: 4–6 fixtures targeting parser-error sites that aren't covered today (e.g. `obtain` missing `where`/`from`, `apply` missing `to`, `rule` with wrong keyword, malformed `assume`).

PR #213, opened 2025-04-22
Parser tests

HalflingHelper:
#208 

Partial coverage of the parser, also noticing that some of the grammar rules in the parser will need to change to match the new lark file.


HalflingHelper:
Looks like this is that weird test thing coming back ?

jsiek:
Ok, I'll go ahead and merge and if the weird error shows up again, I'll look into it.

Issue #217, opened 2025-06-06
Double import error message


Temperz87:
The file:

```
import UInt
import UInt

assert 1 = 1
```

Throws the error:
```
ambiguous overloads:
~/Documents/deduce/lib/UInt.pf:9.3-9.16: (fn UInt -> UInt)
~/Documents/deduce/lib/UInt.pf:9.3-9.16: (fn UInt -> UInt)
```
While a better error might be:
```
Module "UInt" imported twice
```

jsiek:
yes indeed

PR #218, opened 2025-06-07
Double import now has a gooder error message

Temperz87:
The following file (the same as the one in #217):

```
import UInt
import UInt

assert 1 = 1
```
would previously talk about "ambigous overloads" while type checking. Now instead, the it produces the error message:

```
file:2.1-2.12: error, module:
	UInt
was imported twice
```

The way I accomplished this was modifying `type_check_stmt` to take in a set that had the names of already imported modules, then when an import was being typed checked if the name of the imported module was already in the set throw an error. The reason I chose `type_check_stmt` and not `process_declarations` is so I had to modify less functions.

One thing I wanted incorporate into the error message was saying "hey you can safely delete this line" but I couldn't find the wording for it.

### DIFF
--- a/proof_checker.py
+++ b/proof_checker.py
@@ -2585,9 +2585,9 @@ def process_declaration_visibility(decl : Declaration, env: Env, module_chain, d
             ast2.append(new_s)
 
           ast3 = []
-          already_imported = set()
+          already_done_imports = set()
           for s in ast2:
-            new_s = type_check_stmt(s, env, already_imported)
+            new_s = type_check_stmt(s, env, already_done_imports)
             ast3.append(new_s)
 
           for s in ast3:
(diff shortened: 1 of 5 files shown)


proof_checker.py
```diff
@@ -2586,7 +2586,7 @@ def process_declaration_visibility(decl : Declaration, env: Env, module_chain, d
 
           ast3 = []
           for s in ast2:
-            new_s = type_check_stmt(s, env)
+            new_s = type_check_stmt(s, env, None)
```
jsiek, line 2589: How about using an empty set here instead of None?

proof_checker.py
```diff
@@ -2738,6 +2738,10 @@ def type_check_stmt(stmt, env):
       return stmt
   
     case Import(loc, name, ast):
+      if already_done_imports != None:
```
jsiek, line 2741: And delete this line

jsiek:
Looks good now, thanks!

PR #227, opened 2025-09-05
Proof of less_pow_log

HalflingHelper:
Saw this proof while perusing the new library files and gave it a shot.

The lemma `dub_inc_less_dub` might find a better home in `UIntLess` or something so I can move / rename it. I feel like there's a chance it might be applicable to some other proofs?


jsiek:
Excellent, thank you!

PR #240, opened 2025-09-25
Edit distance optimization

mateidragony:
- Removed redundant calls to `closest_keyword`. 
  - What happens now is a first parsing pass with no `closest_keyword` calls. If that parse errors, then we call `closest_keyword` to see if the error was caused by a typo or a more general parse error.
- Added a profiling script which generates a graph of all function calls and the amount of runtime spent on all those calls
- Added a parameter to `parse_proof` to disallow holes on certain calls
  - This is necessary because `cases` calls `parse_proof` on the proof term you are doing case analysis on, and this term should not be a hole.
  - There might be other places we need to set `allow_missing` to False, but I did not do that

> [!Note]
> The error message for `cases_error.pf` is still a little misleading. The issue is that there should be a proof term after the `cases`, but the error message doesn't make that clear.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,9 @@ deduce-release.zip
 # Generated content for site
 gh_pages/pages
 gh_pages/deduce-code
+
+# temp error file
+actual_error.tmp
+
+# Generated deduce profile image
+profile.png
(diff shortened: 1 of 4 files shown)


jsiek:
How much did this optimization improve the runtime on your computer?

mateidragony:
2s -> 0.6s for `example.pf`.

jsiek:
Wow!

Issue #231, opened 2025-09-17
Sums being checked over and over


jsiek:
Deduce is checking `Sums` regardless of whether it has already been checked.

```
python3 /Users/jsiek/deduce/deduce.py /Users/jsiek/scratch.pf
> checking Sums
/Users/jsiek/scratch.pf is valid
```

jsiek:
But this behavior is new to your PR, isn't it.... did you change the behavior to not print empty .thm files?

Temperz87:
This behavior got added [here](https://github.com/jsiek/deduce/blob/dccd536baf63a6c701290ab8f69641c45f6665cd/abstract_syntax.py#L4328) 8 months ago through #86

PR #242, opened 2025-09-26
Remove lib/Sums.pf

Temperz87:
This PR removes the empty `Sums.pf` file inside of `lib`  
Notably however `test/should-validate/fib.pf` imported it so maybe there was a reason it was there?

### DIFF
--- a/test/should-validate/fib.pf
+++ b/test/should-validate/fib.pf
@@ -1,5 +1,4 @@
 import Nat
-import Sums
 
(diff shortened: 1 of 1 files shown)


Temperz87:
With this PR we can also close #231

mateidragony:
Might've been a relic from when we were testing imports that way?

jsiek:
Thanks!

Issue #253, opened 2025-11-06
Clicking "Run" on https://jsiek.github.io/deduce/sandbox.html with empty file produces error


KentShikama:
1. Go to https://jsiek.github.io/deduce/sandbox.html
2. Type nothing into editor - leave the "// Enter deduce code here"
3. Click "Run"
4. The "output" shows `deduce_file() missing 1 required positional argument: 'tracing_functions'`

PR #255, opened 2025-11-06
Site live code bug fix

mateidragony:
Addresses #253 

There was just a missing param to the `deduce_file` call in the api since we added tracing functions.


jsiek:
Thanks!

PR #258, opened 2025-11-14
Custom induction

HalflingHelper:
Code-wise this feels mostly mergeable.

I still need to write tests, and probably see how hard it is to trigger different errors / how good the error messages are.

### DIFF
--- a/Deduce.lark
+++ b/Deduce.lark
@@ -189,6 +189,7 @@ ident: IDENT              -> ident
     | "replace" proof_list "in" proof                    -> rewrite_fact
     | "evaluate"                                         -> eval_goal
     | "evaluate" "in" proof                              -> eval_fact
+    | "simplify" "in" proof                              -> simplify_fact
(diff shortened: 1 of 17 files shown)


HalflingHelper:
Also I still need to make some fixes to induction advice.

HalflingHelper:
I think I feel good about merging this, though there is definitely room for future enhancements to this functionality.

- Known issues / possible solutions
    - Generated proofs fail when the goal is too reductable after expand
      - This is unfixable, because we can't use apply with the goal like this
    - Poor error messages for bad order after "with"
    - If the induction has multiple conditionals, they bundle into one IH
      - Would require assume + multiple haves

  - Possible refactors
    - Use an inductive binding class instead of a dictionary in env
    - Split up generate conjunct body like match induction

  - Add option to choose "inductive" to use at "induction" site.
    - If the type is opaque, then forced to use the custom form
    - Storing multiple versions of custom induction
  - Set is a good example of another opaque thing

PR #264, opened 2025-11-21
Refactoring in to consume-token

HalflingHelper:
Updated the parser and regenerated errors, diff there is just due to consistency of the locations we're using / phrasing / quote changes (I checked them all).

This also has the updated uint_k_induction theorem statement


jsiek:
Thanks Calvin!

Issue #239, opened 2025-09-21
debugger/stepper


jsiek:
want to make it easier to debug Deduce programs

jsiek:
finished

PR #269, opened 2025-12-02
[do not merge] Deduce Debugger

Temperz87:
Adds a gdb like stepper for the functional part of deduce, which turns on when the `--debug` flag is passed in. Currently what's supported is:
- Continue
- Next
- Step Into
- adding break points through the CLI

What needs to be worked in is:
- [ ] Making the CLI user friendly. Namely everything exposed to the end user is ugly
- [ ] Refactoring some parts of the code and adding comments
- [ ] Stepping over theorems automatically
- [ ] The break keyword

What's not supported and won't be included in this PR is:
- Stepping through theorems or proofs in general

I'm also looking for someone to try and break the debugger as I don't know if I've tested every combination of inputs.

### DIFF
--- a/Deduce.lark
+++ b/Deduce.lark
@@ -114,6 +114,7 @@ ident: IDENT              -> ident
     | "if" term "then" term "else" term            -> conditional
     | "if" term "then" term                        -> if_then_formula
     | "%" type
+    | "break" term_hi                              -> break_before
     
 ?term_list:                                        -> empty
     | term                                         -> single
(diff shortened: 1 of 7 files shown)


jsiek:
What does Next do?

jsiek:
This PR is about implementing issue #239

jsiek:
I've picked up where this PR left off in a new PR. Closing this PR.
Thanks for your work on this, it was super helpful!

PR #424, opened 2026-05-10
Fix #365: import using/hiding now applies to prelude (lib/) modules

### LINKED ISSUE #365: import using/hiding no-ops on prelude (lib/) modules
Follow-up to #137 / PR #362.

The `using` / `hiding` clause is a no-op when applied to a module that's also in the auto-prelude (anything in `lib/`). For example:

```
import Nat using operator+
```

is supposed to bring only `+` into scope from `Nat`, but in practice all of `Nat`'s exports are still available because the prelude got there first.

## Why

Auto-prelude is added in [lsp/library.py:190-194](https://github.com/jsiek/deduce/blob/main/lsp/library.py#L190-L194) as `Import(Meta(), name, visibility="private")` for every `lib/*.pf`, prepended before the user's AST. Both prelude and user imports go through `Import.uniquify`, both call `collect_exports` on the imported module's statements, and the env merge is additive — the user's filter only excludes statements its own call processes; it can't remove names the prelude already added.

## Reproducer

```deduce
// foo.pf — auto-prelude imports Nat unfiltered
import Nat using operator+

// `suc` is not in our using list, but it's still in scope:
define _ = suc(0)   // should be an error, but type-checks
```

## Suggested fix

At the prelude-injection site in `lsp/library.py`, scan the user AST first and skip the prelude entry for any module the user is explicitly importing with `using` or `hiding`. The user's import then becomes the only one and the filter takes effect.

This is a small but observable semantic change: user imports would effectively override prelude imports for the same module, which the existing duplicate-detection logic in `proof_checker.py` currently resolves the other way (prelude wins, user's import is dropped). Need to make sure existing tests like `ImportTests.pf` that have `import List` / `import Nat` without filters continue to behave the same — easy if the suppression only triggers when the user's import has a non-`None` `using`/`hiding`.

A `should-validate` (or new `should-error`) test that does `import Nat using <some_name>` and references a filtered-out Nat name would lock in the fix.

jsiek:
## Summary

Closes #365.

When a user file explicitly imports a prelude module with `using` or `hiding`, suppress the prelude's auto-injected unfiltered import of that module so the user's filter actually narrows the visible names.

Before this change, e.g. `import Nat using gcd` was a silent no-op: the prelude brought all of `Nat`'s exports into scope first, and the user's filtered import only contributed names that weren't already there.

## What changed

- [lsp/library.py](lsp/library.py): in `_check_file_impl`, scan the user AST for `Import` nodes with `using` or `hiding` set, and skip the corresponding prelude entries. Untouched modules (and unfiltered user imports) keep the existing behavior.
- New unit test [test/lsp/test_prelude_filter.py](test/lsp/test_prelude_filter.py) exercising four cases against an explicit `Nat`-prelude:
  1. baseline (no user filter): `gcd` is reachable;
  2. `import Nat using equal_refl`: `gcd` is now undefined;
  3. `import Nat using gcd`: `gcd` is reachable;
  4. `import Nat hiding gcd`: `gcd` is undefined.

A `should-error` / `should-validate` fixture would be the more natural test, but the Makefile's `tests-should-validate` target and `test-deduce.py`'s default both run those directories with `--no-stdlib --dir lib`, which disables the auto-prelude — so the bug being fixed can't be exercised through the .pf-fixture harness. The unit test above calls `check_file` directly with an explicit prelude so the filter behaviour is directly observable.

## Test plan

- [x] `python deduce.py tmp/repro.pf` (recursive descent + LALR) errors as expected
- [x] `make tests` (both parsers) green
- [x] `make tests-lib` green
- [x] `python test-deduce.py` (full default suite) green
- [x] `pytest test/lsp/` (808 tests, including the new ones) green
- [x] Verified the new tests fail without the lsp/library.py fix

### DIFF
--- a/lsp/library.py
+++ b/lsp/library.py
@@ -224,8 +224,19 @@ def _check_file_impl(
                 )
 
             if len(prelude) > 0:
+                # If the user file explicitly imports a prelude module
+                # with `using` or `hiding`, suppress the prelude's
+                # unfiltered auto-import for that module so the user's
+                # filter actually takes effect (issue #365).
+                user_filtered = {
+                    s.name for s in ast
+                    if isinstance(s, Import)
+                    and (s.using is not None or s.hiding is not None)
+                }
                 imports = [
-                    Import(Meta(), name, visibility="private") for name in prelude
+                    Import(Meta(), name, visibility="private")
+                    for name in prelude
+                    if name not in user_filtered
                 ]
                 ast = imports + ast
 
--- a/test/lsp/test_prelude_filter.py
+++ b/test/lsp/test_prelude_filter.py
@@ -1,0 +1,55 @@
+"""Acceptance test for issue #365.
+
+When the user file imports a prelude module with `using` or `hiding`,
+the prelude's auto-injected unfiltered import for that same module is
+suppressed so the user's filter actually narrows the visible names.
+"""
+
+from __future__ import annotations
+
+import sys
+from pathlib import Path
+
+REPO_ROOT = Path(__file__).resolve().parents[2]
+sys.path.insert(0, str(REPO_ROOT))
+
+from lsp.library import check_file  # noqa: E402
+
+
+PRELUDE = ("Base", "Option", "NatDefs", "NatAdd", "NatMonus", "NatMult",
+           "NatLess", "NatEvenOdd", "NatDiv", "NatPowLog", "NatSum",
+           "Pair", "Nat")
+
+
+def _check(content: str):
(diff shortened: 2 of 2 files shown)


jsiek:
already done in another PR

jsiek:
Closing — already addressed by #422 (merged ~20 minutes before I opened this). My initial `gh issue list --state open` showed #365 as open, but the issue had been (or was about to be) closed by #422. Apologies for the duplicate.

PR #441, opened 2026-05-12
Fix #406: case_split_at unfolds function-body hypotheses to Or

### LINKED ISSUE #406: MCP: extend `refine_at` templates to cover `≤`, `<`, and other case-split shapes
## Priority: LOW

`refine_at`'s template set is currently shaped for *introduction-style* goals: `true`, conjunction, implication, universal, existential, reducible equation. This misses the most common *elimination-style* goal in arithmetic-heavy proofs: `≤` (which Deduce defines as `< ∨ =` and is almost always destructured by case-splitting on the disjunction).

I wrote out the same `expand operator≤` + `cases` skeleton three times in the proof of [issue #295](https://github.com/jsiek/deduce/issues/295):

```
  have or_eq: a < b or a = b by expand operator≤ in h
  cases or_eq
  case lt: a < b { ? }
  case eq: a = b { ? }
```

## Proposal

Extend `refine_at` with templates for two more shapes when the cursor's goal is a *hypothesis-driven* case-split:

### `≤` (hypothesis: `H: a ≤ b`)
Template:
```
have or_eq: a < b or a = b by expand operator≤ in H
cases or_eq
case lt: a < b { ? }
case eq: a = b { ? }
```

This requires `refine_at` to take an *optional hypothesis* argument rather than only acting on the goal itself — i.e., move closer to the `case_split_at(variable)` shape but for ≤ specifically.

### `<` proven by `≤ and ≠` (or vice versa)
Symmetric template if it ends up being a frequent shape; less sure this one is worth it.

## Why LOW

There's a workable manual recipe (the four-line skeleton above), and the existing `case_split_at` already covers proper `Or`-shaped hypotheses. This is just convenience for the `≤`-disguised-as-Or shape that's pervasive in `lib/UIntLess.pf` style code.

## Related

- Sibling: `refine_at` should return a reason on `null` (low-priority cosmetic).
- The companion `case_split_at` already handles term-level union splits and proof-level `Or` splits; this is the bridge for `≤` which is `<` ∨ `=` definitionally.

jsiek:
## Summary

- Extends `case_split_at(variable=H)` to handle proof hypotheses whose formula is a `Call` to a non-opaque function whose body is a literal `Or` (e.g. UInt's `a ≤ b`, defined as `a < b or a = b`).
- The generated skeleton matches the one in the issue:
  ```
  have or_eq1: (a < b or a = b) by expand operator ≤ in H
    cases or_eq1
      case h1: a < b { ? }
      case h2: a = b { ? }
  ```
- `splittable_vars_at` surfaces those hypotheses too, so the editor's completion candidates include them.

## Design note (vs. issue text)

The issue's title says \"extend `refine_at` templates\", but the body proposal — *an optional hypothesis argument, moving closer to the `case_split_at(variable)` shape* — is what `case_split_at` already is. Lives there rather than as a new `refine_at` arg to avoid a second dual-purpose tool.

## What falls through

- Nat's recursive switch-based `≤` — body isn't a literal `Or`.
- Int's `opaque` `≤` — opaque definitions can't be `expand`ed.

Both produce `None` from `case_split_at`, by design.

## Test plan

- [x] `pytest test/lsp/test_case_split.py test/lsp/test_refine.py test/lsp/test_query.py test/lsp/test_goal_at.py test/lsp/test_mcp_server.py test/lsp/test_lsp_server.py` — 162 passed locally
- [x] Integration probe against the actual UInt prelude: post-edit `check_file` raises only at the two new case-body holes, with goal `a ≤ b`
- [ ] CI: `make tests` (both parsers) + `make tests-lib`

Closes #406

### DIFF
--- a/lsp/query.py
+++ b/lsp/query.py
@@ -1651,6 +1651,10 @@ def _splittable_vars(env) -> tuple:
         if isinstance(binding, ProofBinding):
             if isinstance(binding.formula, Or):
                 seen.add(bname)
+            elif _unfold_call_to_or(binding.formula, env) is not None:
+                # Hypothesis whose definition unfolds to Or, e.g.
+                # UInt's ``a ≤ b`` (issue #406).
+                seen.add(bname)
         elif isinstance(binding, TermBinding):
             ty = binding.typ
             try:
@@ -1682,6 +1686,10 @@ def _case_split_template(name: str, env) -> Optional[str]:
     name ``foo``). Dispatches on the binding's kind:
 
     - ``ProofBinding`` whose formula is ``Or(...)`` -> ``cases``
+    - ``ProofBinding`` whose formula is ``f(a, b)`` where ``f``'s
+      definition unfolds to an ``Or`` (e.g. UInt's ``a ≤ b``, defined
+      as ``a < b or a = b``) -> ``have ... by expand f in <var>`` +
+      ``cases``. Issue #406.
     - ``TermBinding`` whose type resolves to a ``Union`` -> ``switch``
 
     Returns ``None`` for any other shape (built-in atom, function,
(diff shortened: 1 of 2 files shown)


jsiek:
This is too niche

PR #600, opened 2026-05-17
[codex] Type abstract_syntax defs

jsiek:
## Summary

- Enable disallow_untyped_defs and check_untyped_defs for abstract_syntax in mypy config.
- Add return annotations to remaining untyped abstract_syntax.py methods and helpers.
- Fix the small typed-body issues exposed by those annotations.

## Validation

- python3 -m mypy abstract_syntax.py
- python3 -m mypy .
- python3 -m mypy . --no-site-packages

### DIFF
--- a/abstract_syntax.py
+++ b/abstract_syntax.py
@@ -211,7 +211,7 @@ def copy_dict[T](d: dict[str, T]) -> dict[str, T]:
   return {k: v for k, v in d.items()}
 
 
-def maybe_str(o: object | None, default: str = '') -> str:
+def maybe_str(o: Proof | str | None, default: str = '') -> str:
   return str(o) if o is not None else default
 
 def maybe_pretty_print(o: Optional["Proof"], indent: int, default: str = '') -> str:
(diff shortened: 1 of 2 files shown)


abstract_syntax.py
```diff
@@ -211,7 +211,7 @@ def copy_dict[T](d: dict[str, T]) -> dict[str, T]:
   return {k: v for k, v in d.items()}
 
 
-def maybe_str(o: Optional[str], default: str = '') -> str:
+def maybe_str(o: object | None, default: str = '') -> str:
```
jsiek: Is this change necessary?

jsiek: Good catch. The original `Optional[str]` annotation was too narrow once `__str__` bodies are checked: the existing callers pass `Proof` bodies to `maybe_str`, which is why mypy flagged it after the surrounding methods got return annotations. But `object | None` was broader than necessary, so I narrowed it to `Proof | str | None` in beba5d4. `python3 -m mypy abstract_syntax.py` and `python3 -m mypy .` both still pass.

PR #925, opened 2026-06-10
Warn instead of erroring on auto-handled `replace` equations (closes #922)

### LINKED ISSUE #922: Multi-equation `replace A | B` rejected wholesale when only B is auto-handled ("no need for replace ... true")
## Summary

When a `replace` lists several equations with `|` and **one** of them is already
handled by an `auto` rule (e.g. associativity of `*`), Deduce rejects the
**entire** step with

```
no need for replace because this equation is handled automatically
	true
```

even though the *other* equation in the list is genuinely required. The
message tells the user the replace is unnecessary, but deleting it (or the
needed equation) breaks the proof. A beginner who bundles "all the algebra
lemmas I think I need" into one `replace` is told to remove the whole thing,
with no indication of *which* equation was the redundant one.

## Reproduction

`*` on `Nat` is registered associative (so `(a*b)*c` and `a*b*c` are the same
after auto-flattening) but **not** commutative. To prove `(b * a) * c = a * b * c`
you need the commute and the (auto-handled) reassociation:

```deduce
theorem repro: all a:Nat, b:Nat, c:Nat.
  (b * a) * c = a * b * c
proof
  arbitrary a:Nat, b:Nat, c:Nat
  conclude (b * a) * c = a * b * c
      by replace mult_commute[b, a] | mult_assoc[a, b, c].   // rejected
end
```

This fails with:

```
no need for replace because this equation is handled automatically
	true
```

But dropping the auto-handled `mult_assoc` and keeping only the needed
`mult_commute` succeeds:

```deduce
  conclude (b * a) * c = a * b * c
      by replace mult_commute[b, a].   // valid
```

So the `replace` as a whole is *not* unnecessary — only the `mult_assoc[a, b, c]`
entry was.

## Why it's confusing

The wording "no need for replace because this equation is handled
automatically" reads as a verdict on the *whole step*. The natural reaction is
to delete the `replace`, which then fails differently (the sides "differ only
by reordering arguments of `*`"). The two diagnostics point in opposite
directions, so the user bounces between them.

## Suggested fixes (either would help)

1. **Skip the auto-handled equation and apply the rest**, rather than
   rejecting the step. A no-op rewrite in a `|`-list is harmless.
2. If rejecting is preferred, **name the offending equation** and scope the
   message to it, e.g. *"the equation `mult_assoc[a, b, c]` is handled
   automatically — drop it from the replace list"*, instead of implying the
   whole `replace` is unnecessary.

## Context

Found while writing an `examples/` proof (tail-recursive factorial = naive
factorial) as a new user. Deduce 1.x, default recursive-descent parser;
reproduces identically under `--lalr`.

jsiek:
## Summary

Closes [#922](https://github.com/jsiek/deduce/issues/922). When a multi-equation `replace A | B` listed an equation that was already handled by an `auto` rule (e.g. associativity of `*`), Deduce rejected the whole step with `no need for replace because this equation is handled automatically — true`, even when the other equation in the list was genuinely required. The rejection didn't say *which* entry was the redundant one, so beginners bounced between deleting the step (different error) and re-adding it (same rejection).

- Downgrade the rejection to a **warning** that names the offending equation by the user's own syntax (e.g. `` `mult_assoc[a, b, c]` ``), points at that equation's source range, and includes the underlying formula as supporting context.
- Skip the auto-handled entry and keep applying the rest of the `replace`/`simplify` list.
- The filtering lives in a small `_drop_auto_handled_equations` helper called from the three pre-reducing callers (`_check_proof_simplify_fact`, `_check_proof_of_rewrite_goal`, `_check_proof_of_simplify_goal`) so the warning's location is the per-equation range, not the whole tactic step.

## Before / after

Before — the whole step rejected, no hint which entry was redundant:

```
foo.pf:6.10-6.58: no need for replace because this equation is handled automatically
	true
```

After — file validates, warning underlines just the offending entry and names it:

```
foo.pf:6.39-6.58: warning: `mult_assoc[a, b, c]` is handled automatically by an auto rule — drop it from the replace list (it states: (a * b) * c = a * (b * c))
foo.pf is valid
```

## New test category: `test/should-warn/`

Diagnostics that fire on *valid* files had no fixture coverage — `should-error` only locks down rejection text. This PR adds a parallel category:

- `test/should-warn/<file>.pf` must check successfully AND emit warning text that matches the sibling `<file>.pf.warn` fixture (diffed with `--ignore-space-change`, same semantics as `should-error`).
- `test-deduce.py` learns `--warns`, `--regenerate-warns`, and `--generate-warn <path>`. The category joins the default sweep alongside `--passable` / `--errors`, and the parser-equivalence run now also walks `should-warn` so RD/LALR parity is preserved.
- `make tests-should-warn` is the per-category Makefile target.
- `test/should-warn/replace_auto_handled_922.pf` (with `.warn` fixture) covers the issue plus reordered and double-auto-handled variants.

## ⚠️ Manual follow-up: CI matrix entry

My OAuth token lacks `workflow` scope, so I couldn't push the matching matrix entry for `.github/workflows/test_deduce.yml`. Please apply this hunk manually (the rest of the suite still runs `--warns` via the default sweep, but a dedicated CI job is cleaner):

```yaml
          - name: errors
            args: --errors
            requirements: core
+         - name: warns
+           args: --warns
+           requirements: core
          - name: equiv
```

See the [posted comment](https://github.com/jsiek/deduce/pull/925#issuecomment-4672035381) for the same snippet.

## LSP / MCP

`preview_replace_at` previously parsed the old error string to surface a friendly "handled automatically" outcome. With the error gone, that branch is replaced by an `is_true(eq_reduced)` pre-check at the call site, preserving the existing LSP-facing behavior. Prose references to the old wording in `lsp/query.py` and `lsp/mcp_server.py` are refreshed. `pytest test/lsp/test_preview.py` and `test_auto_rules.py` pass (20/20 locally).

## Test plan

- [x] `make static` (ruff + mypy clean)
- [x] `python test-deduce.py` — full default sweep (cli + lib + passable + errors + warns + equiv): all passes, ~205s
- [x] `python test-deduce.py --warns` runs the new category
- [x] Tampering the `.warn` fixture is detected (`warning output diverged: …`); restored fixture passes again
- [x] `python deduce.py` and `python deduce.py --lalr` on the should-warn file both validate with identical warning text
- [x] `pytest test/lsp/test_preview.py test/lsp/test_auto_rules.py` — 20/20 pass

[Claude session](claude://resume?session=05c89185-d189-43b9-801a-8d73d16d2f95) — fallback UUID: `05c89185-d189-43b9-801a-8d73d16d2f95`


jsiek:
The CI matrix update for the new `--warns` category couldn't be pushed — my OAuth token lacks `workflow` scope to modify `.github/workflows/test_deduce.yml`. The local change is staged and tested; please apply this hunk manually (or grant the token `workflow` scope):

```yaml
          - name: errors
            args: --errors
            requirements: core
+         - name: warns
+           args: --warns
+           requirements: core
          - name: equiv
```

`make tests-should-warn` and `python test-deduce.py --warns` already cover this locally and via the default sweep that `make tests` runs.

PR #1020, opened 2026-07-15
RD parser: accept empty-case rule induction/inversion (closes #1018)

### LINKED ISSUE #1018: RD parser rejects empty-case `rule induction`/`rule inversion` while LALR accepts it (parser divergence)
## Bug

`rule induction <pred>` (and `rule inversion <pred>`) with an **empty case
list** is a parse error under the recursive-descent parser but parses fine
under LALR. The two parsers accept different languages here, and the RD
behavior also contradicts the documented design in `Deduce.lark`, which
deliberately allows an empty case list at parse time so the *checker* can
report the missing rule cases with a good diagnostic.

## Repro (in-tree fixture)

```sh
python3 deduce.py --recursive-descent --no-stdlib test/parse/rule_no_case.pf
python3 deduce.py --lalr             --no-stdlib test/parse/rule_no_case.pf
```

`test/parse/rule_no_case.pf` is:

```
theorem t : false
proof
  rule induction P
end
```

## Observed

- **RD:** parse error — `rule induction needs at least one "case" branch`
  (raised in `rec_desc_parser.py:parse_rule_induction`).
- **LALR:** parses; fails later at check time
  (`rule induction: no such predicate 'P'`).

The same split happens for `rule inversion P`.

## Expected

Both parsers accept the empty case list, deferring to the semantic checker,
which already reports the missing cases:
`rule induction: missing cases for rules: ...`
(`checker_proofs.py:_check_rule_induction_or_inversion`, lines ~1307-1313).

## Where

`Deduce.lark` lines ~196-200 explicitly document that an empty case list is
allowed at parse time (for `induction T`, `cases p`, `switch x for {}`, and
`rule induction P`) so semantic checks produce the error. RD's
`parse_induction` follows that, but `parse_rule_induction`
(`rec_desc_parser.py:1379`) raises a parse error when `not cases`, diverging
from both LALR and RD's own `induction` handling.

Found via a local round-trip/equivalence audit over `test/parse/` +
`test/should-error/` (the corpora that `--equiv-full` deliberately excludes).

Refs #473.

jsiek:
## Summary

Fixes the parser divergence in #1018: the recursive-descent parser rejected
`rule induction <pred>` / `rule inversion <pred>` with an **empty case list**
as a parse error, while the LALR parser accepted it and deferred to the
semantic checker.

The RD behavior contradicted:
- **LALR**, which parses the empty case list and fails later at check time.
- **RD's own `parse_induction`**, which already tolerates an empty case list.
- **`Deduce.lark`** (lines ~196-200), which deliberately allows an empty case
  list at parse time so the checker can report the missing rule cases with a
  good diagnostic (`checker_proofs.py:_check_rule_induction_or_inversion`).

## Change

- `rec_desc_parser.py`: drop the RD-only `rule … needs at least one "case"
  branch` error in `parse_rule_induction`. Both parsers now defer to the
  checker, which reports `rule induction/inversion: missing cases for rules: …`.
- Test fixtures:
  - Removed `test/parse/rule_no_case.pf` (+ `.err`), which locked in the buggy
    RD-only parse diagnostic and is no longer a parse error.
  - Added `test/should-error/rule_induction_empty_cases.pf` and
    `rule_inversion_empty_cases.pf` (+ `.err`), which use a real predicate so an
    empty case list reaches the documented "missing cases" checker path.
  - Added both new fixtures to `PARSER_EQUIV_FILES` in `test-deduce.py` so the
    RD/LALR ASTs are compared, guarding against this divergence recurring.

## Verification

```
python3 test-deduce.py            # full default suite — All tests passed
python3 test-deduce.py --equiv    # RD vs LALR AST parity — All tests passed
python3 test-deduce.py --errors   # should-error fixtures — All tests passed
python3 test-deduce.py --parser   # test/parse fixtures — All tests passed
```

(ruff/mypy are not installed in this container; `make static` runs them in CI.)

Closes #1018.

Resume this session on ginger: `~/deduce-runner/resume.sh 65dc7bed-cb01-41f0-87ab-b86edb91c7cd routine/20260715-100202`


jsiek:
This issue is already fixed.

Issue #1049, opened 2026-07-17
CI is ~13 min, bottlenecked on the unsharded 'passable' job; trim to ~3 min without losing coverage


jsiek:
## Summary

CI (`Run Tests`) wall-clock is ~13 minutes and is dominated almost entirely by
a single unsharded job. We should trim the test time to a target of **~3
minutes** without meaningfully losing coverage.

## Evidence

Per-job timing for run
[29581564644](https://github.com/jsiek/deduce/actions/runs/29581564644)
(~13m19s total; all jobs start together at 12:48:09):

| job | duration |
|-----|----------|
| `regression (passable)` | **~13m19s** ← the entire critical path |
| `regression (equiv)` | ~4m58s |
| `regression (lib)` | ~3m36s |
| `regression (examples)` | ~2m39s |
| `regression (errors)` | ~2m25s |
| `regression (site)` | ~1m57s |
| `regression (parser)` | ~1m56s |
| `regression (warns)` | ~1m49s |
| `static checks` | ~0m30s |

The matrix already runs these in parallel, so total wall-clock ≈ the slowest
job. Everything except `passable` already finishes in ≤5 min; most in ≤3.

## Root cause

`regression (passable)` runs `python test-deduce.py --passable --cli`, which
checks all **199** `test/should-validate/*.pf` files **under both parsers**
(RD + LALR) — 398 checks — and each check prepends the full **41-file stdlib
prelude**. Locally on 24 cores this is ~224s ("should-validate × 2 parsers");
on a 2–4 core GitHub runner it stretches to ~13 min because it is a single
unsharded job with little parallelism to exploit.

## Proposed directions (aim: ~3 min wall-clock)

Roughly in order of leverage; a combination likely gets us there.

1. **Shard `passable` across the matrix.** Add a `--shard i/n` option to
   `test-deduce.py` and split should-validate into ~4–5 parallel matrix legs.
   This alone should take the long pole from ~13m to ~3m.
2. **Don't run every should-validate file under both parsers.** RD/LALR AST
   parity is already the job of `--equiv` (curated corpus). Let `--passable`
   run a single parser (the default RD) for validity, and rely on `--equiv`
   for cross-parser agreement — roughly halving `passable`.
3. **Trim the other >3-min jobs too** (`equiv` ~5m, `lib` ~3.5m) — shard or
   curate so no single leg exceeds the target.
4. **Cut redundant prelude re-checking.** Each should-validate file re-checks
[...]

Issue #1025, opened 2026-07-15
Represent UInt literals in compact binary form end-to-end (deferred from #1021)


jsiek:
## Goal

Represent `UInt` numeric literals in their compact binary form end-to-end, so
a literal's cost is `O(log value)` instead of `O(value)` — removing the
recursion ceiling that #1021 / #1023 worked around with a parse-time cap
(`flags.MAX_LITERAL = 4000`) and a raised `flags.RECURSION_LIMIT`.

This is the follow-up tracked by the "Root cause" note in #1021: the interim
fix (PR #1023) makes oversized literals fail with a clean, located error, but
the underlying representation is still unary.

## Background

- A `UInt` literal `N` currently desugars (in both parsers, via
  `mkUIntLit`) to `fromNat(lit(suc^N(zero)))` — a **unary** `Nat` of depth
  `N` wrapped by `fromNat`. A `Nat` literal `ℕN` desugars to `lit(suc^N(zero))`.
- Every tree-walking pass (uniquify, type-check, evaluation, and the
  literal recognizers `natToInt` / `uintToInt` /
  `formulas_equal_modulo_numeric_literals`) recurses to that depth, so the
  practical ceiling is ~5000 (CPython's C-stack guard). Measured cost is a
  clean ~5 recursion frames per unit of value.
- `abstract_syntax/literals.py:intToUInt` already builds the compact binary
  (`bzero` / `inc_dub` / `dub_inc`) representation and, since #1023, does so
  in `O(log n)` depth — but **nothing calls it**.

## The tension: who may name the representation?

`intToUInt` is unused because `Binary`'s constructors are **`private`** to the
UInt module:

```
private union Binary {
  bzero
  dub_inc(Binary)
  inc_dub(Binary)
}
```

The library hides the representation on purpose (`lib/UIntDefs.pf`: *"The
representation is hidden from clients"*) so user code can't pattern-match on or
construct binary internals — clients see only arithmetic, order, and `toNat`.

But the parsers work by emitting pre-uniquify `Var(name)` nodes that `uniquify`
later resolves against the importing module's environment. A user file that
imports `UInt` does **not** have `bzero` / `inc_dub` / `dub_inc` in scope
(verified: `assert toNat(inc_dub(bzero)) = ℕ3` errors with *"Add `import UInt`
to use unsigned integers"*), so the parser cannot emit them today. And
uniquified names carry an unstable `base.scopeN` suffix (`core.py:generate_name`),
[...]

PR #1059, opened 2026-07-17
Fast-path UInt literal evaluation under eval_all (closes #1050)

### LINKED ISSUE #1050: UInt arithmetic on large literals evaluates in O(n) (unary representation): 4000=4000 takes ~37s
## Summary

Runtime *evaluation* of arithmetic/comparison on large `UInt` literals is
pathologically slow because numeric literals are represented in unary. A single
`assert` on values near `MAX_LITERAL` (4000) takes tens of seconds, while merely
*type-checking* the same literal is ~0.5s.

## Evidence (measured via `check_file`, RD parser, warm prelude)

| expression | time |
|------------|------|
| `define big : UInt = 4000` (type-check only) | 0.51s |
| `define big : Nat = ℕ4000` (type-check only) | 0.27s |
| `assert ℕ3000 = ℕ3000` (Nat eval) | 0.39s |
| `assert 3000 = 3000` (UInt eval) | 21.8s |
| `assert 3000 < 4000` (UInt eval) | 30.4s |
| `assert 4000 = 4000` (UInt eval) | 37.3s |
| `assert 1500 + 1500 = 3000` (UInt eval) | 16.1s |
| `assert 2000 + 2000 = 4000` (UInt eval) | 29.2s |

Note the ~50–100× gap between the `Nat` (`ℕ`-prefixed) eval path and the `UInt`
decimal-literal eval path for equal magnitudes, and between type-checking a large
`UInt` literal (cheap) and evaluating one (expensive).

## Impact

`test/should-validate/large_numeric_literals.pf` (a #1021 regression test) was a
single file costing ~107s **per parser** — ~214s of the ~224s local
`should-validate` runtime, and effectively the entire ~13-min CI bottleneck
(#1049). It was trimmed to keep the type-check coverage (the actual #1021
regression) plus a cheap `Nat`-path eval, dropping the expensive large-`UInt`
runtime-eval asserts. This issue tracks the underlying perf so that coverage can
be restored cheaply once literals evaluate efficiently.

## Root cause / fix direction

Unary literal representation makes `UInt` arithmetic O(n) (or worse) in the value.
This is the symptom that the compact binary `UInt` literal representation in
**#1025** is meant to remove. `Refs #1025`, `Refs #1049`.

## Where found

Surfaced while profiling the CI bottleneck for #1049 (per-file timing of
`test/should-validate`).

jsiek:
## Summary

Closes #1050.

Runtime evaluation of `UInt` arithmetic/comparison on large literals was
pathologically slow. A `UInt` literal enters the evaluator as
`fromNat(<unary Nat of depth N>)`, whose `opaque` unfolding is O(N), and the
recursive `Binary` division operator is O(N/M). Under `assert`/`evaluate`
(which set `eval_all`) these unfold through the full tree-rewriter, so
`assert 4000 = 4000` took ~37s and the `large_numeric_literals.pf` regression
test cost ~107s **per parser** — effectively the entire CI bottleneck tracked
in #1049.

## Approach

Rather than block on the full binary-literal migration (#1025, which changes the
literal *AST shape* and would break `UInt` proof automation globally), this adds
an evaluation fast path that keeps the shape and only changes reduction speed —
directly analogous to the existing Nat `try_fast_lit_nat_arith`.

`try_fast_uint_arith` (in `abstract_syntax/literals.py`, called from
`Call.reduce`) recognizes concrete `UInt` literal operands and computes the
result with Python ints, emitting the compact `Binary` value directly:

- `fromNat(<concrete Nat>)` → the compact `Binary` value in O(log N) (this is
  the dominant win; once operands are compact, `=`/`<`/`+` are already O(log N)).
- `+`, `*`, `∸`, `/`, `%`, `<`, `≤` on two concrete `UInt` literals.
- `=` is left to the generic structural-equality path (operands are already
  compact after `fromNat` is fast).

It fires **only under `eval_all`** (`assert`/`evaluate`), where the
`fromNat(lit(...))`/`Binary`-operator shape is being reduced away regardless, so
the result is structurally identical to what the slow opaque unfolding produces
and symbolic proofs are untouched. It returns `None` unless every operand is a
concrete `UInt` literal, so symbolic terms fall through to normal reduction.
`n / 0 = 0` and `n % 0 = n` match `lib/UIntDiv.pf`.

**Not hijacking user overloads** (per the Codex review): matching by base name
alone would evaluate a same-named user overload (e.g. `fun fromNat(n:Nat) -> Nat`)
as the built-in UInt conversion. Value-producing operations therefore also
require the call's result type to be the UInt library's `private` `Binary`
representation — unspoofable by user code — so a user `fromNat` returning `Nat`
(or any op not returning UInt) is left to normal reduction. Comparisons return
`bool`, so they keep the operand-type guard (both operands must be genuine UInt
values). Covered by `test/should-validate/uint_fast_path_overload.pf`, which
fails on the pre-fix logic.

## Results

| expression | before | after |
|------------|--------|-------|
| `assert 4000 = 4000` | ~37s | <0.1s |
| `assert 3000 < 4000` | ~30s | <0.1s |
| `assert 2000 + 2000 = 4000` | ~29s | <0.1s |
| `large_numeric_literals.pf` (whole file) | ~107s/parser | ~14s/parser |

The trimmed UInt runtime-eval coverage is restored in
`test/should-validate/large_numeric_literals.pf` (equality, ordering, and each
arithmetic operator near `MAX_LITERAL`), now that it is cheap.

## Testing

- `python3 test-deduce.py` (full default harness: lib + should-validate ×2
  parsers + should-error + should-warn + parser equivalence + round-trip) — all
  pass.
- `python3 -m ruff check .` and `python3 -m mypy .` — clean.
- Correctness spot-checks against hand-computed values for `=`, `<`, `≤`, `+`,
  `∸`, `/`, `%`, including saturating monus, div/mod by zero, and negative
  (must-fail) assertions.

Resume this session on ginger: `~/deduce-runner/resume.sh 97a39018-ccaf-4e38-b5e2-d5b566d0f928 routine/20260717-220201`

### DIFF
--- a/abstract_syntax/literals.py
+++ b/abstract_syntax/literals.py
@@ -527,6 +527,57 @@ def try_fast_lit_nat_arith(loc: Meta, rator: Term, args: list[Term],
                    sname=suc_name or 'suc')
   return Call(loc, ty, _resolved_or_var(loc, None, lit_name), [inner])
 
+def _mk_uint_from_int(loc: Meta, n: int, ty: Type | None, env: Env) -> Term | None:
+  # Build the compact ``Binary`` value for ``n`` using the environment's
+  # uniquified constructor names, so the result is structurally identical
+  # to what the opaque recursive reduction would produce.
+  bz = env.base_to_unique('bzero')
+  di = env.base_to_unique('dub_inc')
+  idd = env.base_to_unique('inc_dub')
+  if bz is None or di is None or idd is None:
+    return None
+  return intToUInt(loc, n, bzero=bz, dubinc=di, incdub=idd, uint_ty=ty)
+
+def try_fast_uint_arith(loc: Meta, rator: Term, args: list[Term],
+                        ty: Type | None, env: Env) -> Term | None:
+  # Full-evaluation (``assert``/``evaluate``) fast path for UInt.  A UInt
+  # literal enters the evaluator as ``fromNat(<unary Nat of depth N>)``,
+  # whose opaque unfolding is O(N); and the recursive ``Binary`` division
+  # operator is O(N/M).  Both make ``assert`` on large literals take tens
+  # of seconds (issue #1050).  Here we compute the result with Python ints
+  # and emit the compact ``Binary`` value directly.  Returns None unless
+  # every operand is a concrete UInt literal, so symbolic terms fall
+  # through to normal reduction.  Only sound under ``eval_all``, where the
+  # ``fromNat``/``Binary``-operator shape that proof automation keys on is
+  # already being reduced away.
+  if not isinstance(rator, (Var, OverloadedVar, ResolvedVar)):
+    return None
+  op = base_name(rator.get_name())
+  if op == 'fromNat' and len(args) == 1 and isNat(args[0]):
+    return _mk_uint_from_int(loc, natToInt(args[0]), ty, env)
+  if op not in ('+', '*', '∸', '/', '%', '<', '≤'):
+    return None
+  if len(args) != 2 or not (isUInt(args[0]) and isUInt(args[1])):
+    return None
+  a = uintToInt(args[0])
+  b = uintToInt(args[1])
+  if op == '<':
+    return Bool(loc, ty, a < b)
+  if op == '≤':
+    return Bool(loc, ty, a <= b)
+  if op == '+':
+    result_int = a + b
+  elif op == '*':
+    result_int = a * b
+  elif op == '∸':
+    result_int = max(0, a - b)
+  elif op == '/':
+    # UInt division truncates and defines n / 0 = 0 (lib/UIntDiv.pf).
+    result_int = a // b if b != 0 else 0
+  else:  # '%' : n % m = n ∸ (n / m) * m, so n % 0 = n (lib/UIntDiv.pf).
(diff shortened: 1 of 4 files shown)


jsiek:
Thanks — good catch. Fixed in 3b8ef59.

The fast path matched operators by base name only, so a same-named user overload could be hijacked. It now anchors on the **result type**: value-producing operations (`fromNat`, `+`, `*`, `∸`, `/`, `%`) require the call's result type to be the UInt library's private `Binary` representation. `Binary` is `private` to the UInt module, so no user file can name it — a user `fun fromNat(n:Nat) -> Nat` (result type `Nat`) is therefore left to normal reduction. Comparisons return `bool`, so they keep the operand-type guard (both operands must be genuine UInt values built from the private UInt constructors), which is the same guard the existing Nat fast path uses.

Added `test/should-validate/uint_fast_path_overload.pf` as a regression test — it defines `fun fromNat(n:Nat) { n + ℕ1 }`, uses it at expected type `Nat`, and asserts the result is `ℕ6`. On the pre-fix logic this failed with `5 ≠ 6` (the user `fromNat` collapsed to UInt 5); with the result-type anchor it passes.

Issue #473, opened 2026-05-12
Dual-parser equivalence testing + sync Reference.md grammar with LALR


jsiek:
## Status after #893

Done:

- Migrated more Reference.md operator and literal grammar snippets to strict checked `deduce-grammar` blocks.
- Used Deduce.lark precedence nonterminals for arithmetic, logical, comparison, membership, and literal snippets.
- Corrected stale documented ASCII aliases for multiset addition and function composition from `[+]`/`[o]` to `.+.`/`.o.`.
- Extended `gh_pages/scripts/reference_grammar.py` so precedence-level entries can be subset-checked one syntax form at a time.

Still left to do:

- Continue migrating remaining Reference.md grammar fragments.
- Decide whether and how much of `test/should-error/` should participate in parser equivalence checks.
- Expand round-trip coverage after improving the pretty-printer for currently non-parse-preserving proof/declaration forms.

## Status after #575

Done:

- Added representative parser round-trip coverage to `test-deduce.py --equiv`.
- The new check parses selected accepted programs with both parsers, pretty-prints each AST, re-parses that generated source with both parsers, and compares canonical ASTs.
- Kept the round-trip corpus curated because the current pretty-printer is not parse-preserving for the full lib/should-validate corpus yet.
- Made `gh_pages/scripts/convert.py` pass mypy under the required `make tests` static gate.

Still left to do:

- Decide whether and how much of `test/should-error/` should participate in parser equivalence checks.
- Migrate more `Reference.md` grammar fragments to strict `deduce-grammar` blocks, marking intentional simplifications before strict comparison.
- Expand round-trip coverage after improving the pretty-printer for currently non-parse-preserving proof/declaration forms.

## Status after #554

Done:

- Added `gh_pages/scripts/reference_grammar.py` to compare strict `deduce-grammar` blocks in `gh_pages/doc/Reference.md` against `Deduce.lark`.
- Migrated the Type, Type List, Type Parameters, and Visibility grammar snippets to strict checked blocks.
- Wired the checker into `make tests-tokens`, the main test workflow, and the static site workflow.

Still left to do:

- Decide whether and how much of `test/should-error/` should participate in parser equivalence checks.
- Add round-trip property tests: pretty-print an AST, re-parse with each parser, and assert equivalence.
- Migrate more `Reference.md` grammar fragments to strict `deduce-grammar` blocks, marking intentional simplifications before strict comparison.

## Status after #530

Done:

[...]

PR #1060, opened 2026-07-17
Fix RD and/or precedence to match LALR left-to-right grouping (refs #473)

jsiek:
## Summary

Fixes an RD/LALR parser divergence on the relative precedence and associativity of `and`/`or`, surfaced while working the parser-equivalence portion of #473.

An in-process RD-vs-LALR snippet sweep over grammar forms **absent from the curated `--equiv` corpus** found:

```
define d = true and false or true
```

- recursive-descent (the default parser): `true and (false or true)`
- LALR: `(true and false) or true`

The two parsers disagreed on the accepted grouping: RD made `or` bind *tighter* than `and`, and made both right-associative.

## Root cause

`rec_desc_parser.py:parse_term_logic` parsed the right operand of `and`/`or` by recursing back into `parse_term_logic`, which consumes the entire remaining logical expression. That yields right-associativity with `or` grouping inside `and`.

`Deduce.lark` defines `and`/`or`/`:` at a single left-recursive `logical_term` level:

```
?logical_term: logical_term "and" sep_term  -> and_formula
    | logical_term "or" sep_term            -> or_formula
    | logical_term ":" type                 -> annote_type
```

and the operator-precedence table in `gh_pages/doc/Reference.md` documents them as one level "read left to right" — i.e. `a and b or c` is `(a and b) or c`. RD was the incorrect parser, and it is the default.

## Fix

Parse the right operand at the next-tighter level (`parse_term_sep`) inside the `while` loop, making the level left-associative and matching both the grammar and the docs. Pure `a and b and c` / `a or b or c` chains are unchanged (the `extract_and`/`extract_or` flattening already produced identical flat ASTs); only mixed `and`/`or` and the `and`/`or`/`:` interaction change, and no file in the corpus exercised those (the full 508-file RD-vs-LALR AST sweep was already clean).

## Test coverage

Extended `test/should-validate/operator_precedence_rd_lalr.pf` (already in the `--equiv` corpus) with two theorems that only validate under the correct left-to-right grouping:

- `all P:bool, Q:bool. (P or Q and false) = false` — `(P or Q) and false` is always false; the old RD grouping `P or (Q and false)` reduces to `P`.
- `all P:bool, Q:bool. (P and Q or true) = true` — `(P and Q) or true` is always true; the old RD grouping reduces to `P`.

## Verification

- `python3 test-deduce.py` — all categories pass (both parsers).
- `python3 test-deduce.py --equiv` — parser-equivalence corpus green.
- `python3 gh_pages/scripts/keywords.py` and `reference_grammar.py` — green.
- ruff/mypy not run here (not installed in the container); the change is a one-line call-site swap plus comments and a `.pf` fixture.

## Resuming

Since this only advances the parser-equivalence portion of the umbrella issue #473, it uses `Refs`, not a closing keyword.

Refs #473.

Resume this session on ginger: `~/deduce-runner/resume.sh 3702379f-dcff-4ac3-a87b-8d4fa8cfbba2 routine/20260717-230202`

### DIFF
--- a/rec_desc_parser.py
+++ b/rec_desc_parser.py
@@ -642,7 +642,12 @@ def parse_term_logic() -> Term:
                                  or current_token().type == 'OR'):
     opr = current_token().type
     advance()
-    right = parse_term_logic()
+    # Left-associative at a single `and`/`or` precedence level, matching the
+    # left-recursive `logical_term` rule in Deduce.lark. Recursing into
+    # `parse_term_logic` here would instead group `a and b or c` as
+    # `a and (b or c)`, disagreeing with the LALR parser and the documented
+    # precedence (Reference.md: `and`/`or`/`:` are one level, read left to right).
+    right = parse_term_sep()
     if opr == 'AND':
       term = And(meta_from_tokens(token, previous_token()), None,
                  extract_and(term) + extract_and(right))
--- a/rec_desc_parser.py
+++ b/rec_desc_parser.py
@@ -638,28 +638,28 @@ def parse_term_sep() -> Term:
 def parse_term_logic() -> Term:
   token = current_token()
   term = parse_term_sep()
-  while (not end_of_file()) and (current_token().type == 'AND'
-                                 or current_token().type == 'OR'):
+  # `and`, `or`, and the type annotation `:` share one left-associative
+  # precedence level, matching the left-recursive `logical_term` rule in
+  # Deduce.lark and the documented precedence (Reference.md: read left to
+  # right). Handling any of them by recursing into `parse_term_logic`, or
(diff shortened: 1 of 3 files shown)


jsiek:
Good catch — fixed in acbfaf2 (rebased to 8dc4aab). `:` is at the same left-recursive `logical_term` level as `and`/`or` in Deduce.lark, so it's now handled as a third case *inside* the same left-associative loop rather than in a post-loop block. `P or Q : bool and false` now parses as `((P or Q):bool) and false` under both parsers (it previously errored under RD with `and false` unconsumed). Added a discriminating regression theorem `annotation_same_level_as_and` to the `operator_precedence_rd_lalr.pf` equiv fixture; `--equiv` and the full harness are green.

PR #1063, opened 2026-07-18
Render UInt zero value as 0 instead of bzero (closes #1062)

### LINKED ISSUE #1062: UInt zero value prints as `bzero` instead of `0`
## One-sentence summary

A `UInt` value that reduces to zero displays as the internal constructor name `bzero` instead of the decimal literal `0`, even though every nonzero `UInt` value prints as a decimal.

## Repro

```sh
cat > /tmp/bzero.pf <<'PF'
import UInt
print 0
print 1
print 2
assert 3 % 3 = 1
PF
python3 deduce.py /tmp/bzero.pf
```

## Observed vs. expected

Observed:

```
bzero
1
2
/tmp/bzero.pf:5.1-5.17: assertion failed:
	bzero ≠ 1
```

Expected:

```
0
1
2
/tmp/bzero.pf:5.1-5.17: assertion failed:
	0 ≠ 1
```

`print 1` / `print 2` already render as decimals, so the zero case is inconsistent, and the raw internal name leaks into `print` output and error/goal messages.

## Root cause

`UInt` values use the compact binary representation (`bzero` / `inc_dub` / `dub_inc`). A nonzero value is a `Call` (`inc_dub(...)` / `dub_inc(...)`), so `Call.__str__` in `abstract_syntax/terms.py:1196` hits its `isUInt(self) and not get_verbose()` branch and prints `str(uintToInt(self))` — a decimal. But `UInt` zero is the bare nullary constructor `bzero`, a `ResolvedVar`, which never reaches `Call.__str__`; `ResolvedVar.__str__` (`abstract_syntax/terms.py:708`) has no `bzero`→`0` case (it only special-cases the list-nil `empty`→`[]`), so it falls through to `name2str(self.name)` and prints `bzero`.

## Where found

Surfaced while auditing `Int`/`UInt` stdlib coverage for #720: `print 0 % 0` and `print 3 % 3` printed `bzero`. Independent of the #1025 literal-*parsing* migration — this is a defect in the display of the already-live binary *value* representation. Refs #1025 (its scope notes call out that the `Call.__str__` pretty-print path must round-trip binary values back to a decimal; this closes the `bzero` gap in that path).

jsiek:
## What

A `UInt` value that reduces to zero displayed as the internal binary
constructor name `bzero` instead of the decimal literal `0`, in both `print`
output and error/goal messages:

```
$ cat bzero.pf
import UInt
print 0
print 1
assert 3 % 3 = 1
$ python3 deduce.py bzero.pf     # before
bzero
1
bzero.pf:4.1-4.17: assertion failed:
	bzero ≠ 1
```

Every *nonzero* `UInt` value already printed as a decimal, so zero was the odd
one out and leaked an internal name to users.

## Why

`UInt` values use the compact binary representation (`bzero` / `inc_dub` /
`dub_inc`). A nonzero value is a `Call` (`inc_dub(...)` / `dub_inc(...)`), so
`Call.__str__` (`abstract_syntax/terms.py`) hits its `isUInt(self) and not
get_verbose()` branch and prints `str(uintToInt(self))`. But `UInt` zero is the
bare nullary constructor `bzero`, a `ResolvedVar`, which never reaches
`Call.__str__`; `ResolvedVar.__str__` had no `bzero`→`0` case (only the
list-nil `empty`→`[]` sugar), so it fell through to `name2str` and printed
`bzero`.

## Change

- Give `ResolvedVar.__str__` a `bzero`→`"0"` case, gated on the same
  `not get_verbose()` condition the `Call` UInt branch uses. All `UInt` values
  now display as decimals in the default and `--unique-names` modes and stay
  raw only under `--verbose` (which is the mode that intentionally shows binary
  internals).
- Add `test/unit/test_uint_display.py` locking in the display of `0`, nonzero
  values, the `--verbose` raw form, and the `--unique-names` case. (This ran as
  `make tests-unit`; verified it fails on the two zero cases without the fix.)

After:

```
$ python3 deduce.py bzero.pf
0
1
bzero.pf:4.1-4.17: assertion failed:
	0 ≠ 1
```

## Testing

- `python3 -m ruff check .` — clean
- `python3 -m mypy .` — clean (54 files)
- `python3 -m pytest test/unit/` — 566 passed
- `python3 test-deduce.py` — all passed (lib + should-validate ×2 parsers +
  should-error + should-warn + parser equivalence)

## Scope note

This is a defect in the display of the already-live binary *value*
representation and is independent of the larger #1025 literal-*parsing*
migration; it closes the `bzero` gap in the pretty-print path that #1025's
scope notes call out. `Refs #1025`.

Closes #1062
Refs #1025

Resume this session on ginger: `~/deduce-runner/resume.sh 8f7c8bb8-4fc8-4133-a78a-df5c1df8dcb7 routine/20260718-010202`

### DIFF
--- a/abstract_syntax/terms.py
+++ b/abstract_syntax/terms.py
@@ -325,6 +325,13 @@ class PatternCons(Pattern):
     return PatternCons(self.location, self.constructor, params)
 
   def __str__(self) -> str:
+      # A pattern's constructor must print as a constructor, never as the
+      # decimal value-sugar `ResolvedVar.__str__` gives UInt zero: `case 0`
+      # would reparse as the `Nat` `zero` pattern, not the `Binary` `bzero`
+      # constructor (unlike `empty`->`[]`, which does reparse as the nil
+      # pattern).
+      if isBZero(self.constructor):
+        return 'bzero'
       if len(self.parameters) > 0:
         ctor = applied_head_str(self.constructor) or str(self.constructor)
         return ctor \
--- a/abstract_syntax/terms.py
+++ b/abstract_syntax/terms.py
@@ -44,6 +44,7 @@ if TYPE_CHECKING:
         getSuc,
         getZero,
         intToNat,
+        isBZero,
         isDeduceInt,
         isLitNat,
         isLitUInt,
@@ -703,6 +704,11 @@ class ResolvedVar(VarRef):
   def __str__(self) -> str:
     if base_name(self.name) == 'empty' and not get_unique_names() and not get_verbose():
       return '[]'
+    elif isBZero(self) and not get_verbose():
+      # UInt zero is the value `bzero`; render it as the decimal literal `0`,
+      # matching the `Call.__str__` UInt branch that already prints nonzero
+      # binary values (inc_dub/dub_inc trees) back as decimals.
+      return '0'
     elif get_unique_names():
       return name2str(self.name) + '{' + self.name + '}'
     elif is_var_operator(self):
--- a/test/unit/test_uint_display.py
+++ b/test/unit/test_uint_display.py
@@ -43,3 +43,12 @@ def test_uint_zero_prints_decimal_under_unique_names() -> None:
     # decimals; zero must be consistent with them.
     set_unique_names(True)
     assert str(ast.intToUInt(Meta(), 0)) == "0"
+
+
+def test_bzero_pattern_constructor_stays_bzero() -> None:
+    # The value-sugar must not leak into pattern/constructor position:
+    # `case 0` would reparse as the Nat `zero` pattern, not the Binary
+    # `bzero` constructor. A resolved `bzero` pattern constructor (as the
+    # induction/predicate machinery builds) must print as `bzero`.
+    pat = ast.PatternCons(Meta(), ast.ResolvedVar(Meta(), None, "bzero"), [])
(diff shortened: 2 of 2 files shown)


jsiek:
Thanks — good catch on the constructor-position risk. Addressed in 5982dbc (rebased to 8f0b5c9).

The concrete `.thm`/signature path you cited is actually safe today: parser-built UInt switch patterns keep a `Var` constructor (never reassigned to `ResolvedVar`), so `PatternCons.__str__` printed `case bzero` correctly and my `ResolvedVar.__str__` change never fired there — verified by regenerating `lib/UIntDefs.thm` (still `case bzero`, `toNat(bzero) = ...`).

But the underlying concern is valid: the induction/predicate machinery (`checker_predicates.py`) *does* build `PatternCons` with a `ResolvedVar` constructor, so the value-sugar could leak into constructor position, and `case 0` reparses as the `Nat` `zero` pattern rather than the `Binary` `bzero` constructor — unlike `empty`→`[]`, which reparses as the nil pattern. So I've guarded `PatternCons.__str__` to print a `bzero` constructor as its raw name regardless of the value-sugar. Added a unit test (`test_bzero_pattern_constructor_stays_bzero`) locking that in, so a resolved `bzero` **value** prints `0` while a `bzero` **pattern constructor** prints `bzero`.

PR #1067, opened 2026-07-18
Respect private visibility on postulates when exporting (closes #1065)

### LINKED ISSUE #1065: private postulate is silently exported across modules (visibility ignored)
## One-sentence summary

`private postulate` is silently exported to importing modules — `Postulate.collect_exports` ignores visibility entirely, so a postulate declared `private` in one module is usable from another, unlike every other module-private declaration.

## Is this intentional? If so, close.

Postulates may be meant as always-global axioms. But the `private` keyword parses on a postulate (it sets `visibility = 'private'` and even prints via `visibility_prefix`), so silently ignoring it is at best surprising. If postulates are intended to always be public, `private postulate` should probably be rejected at parse/check time rather than accepted-and-ignored.

## Repro

```sh
mkdir -p /tmp/pt
cat > /tmp/pt/ModA.pf <<'PF'
module ModA
private postulate secret_fact: true = false
PF
cat > /tmp/pt/useit.pf <<'PF'
import ModA
theorem uses_secret: true = false proof secret_fact end
PF
python3 deduce.py --no-stdlib --dir /tmp/pt /tmp/pt/useit.pf
```

## Observed vs. expected

- Observed: `/tmp/pt/useit.pf is valid` — the private postulate `secret_fact` is accessible from another module (and here even lets an importer "prove" `true = false`).
- Expected: parity with the module-private `lemma`, which is correctly hidden. The analogous `lemma` case:

  ```
  module ModC
  lemma hidden_lemma: true = true proof . end
  ```

  used from another module errors with:

  ```
  'hidden_lemma' is declared as a `lemma` in module ModC (...); lemmas are
  module-private and not accessible from other modules.
  ```

  A `private postulate` should either be hidden the same way, or be rejected outright.

## Root cause

`abstract_syntax/declarations.py`, `Postulate.collect_exports`:

```python
def collect_exports(self, export_env: UniquifyEnv, importing_module: str) -> None:
    export_env[base_name(self.name)] = [self.name]
```

It exports unconditionally — no `self.visibility == 'private' and importing_module != get_current_module()` guard, which every name-exporting `Declaration` (Union, TypeAlias, Define, RecFun, …) has. So `private` on a postulate has no effect on export.

## Where found

Noticed while deduplicating the `collect_exports` methods for #813 (the `Postulate` override was the one name-exporting declaration whose body did not fit the shared private-visibility pattern). File path: `abstract_syntax/declarations.py`.

jsiek:
## Summary

`private postulate` was silently exported to importing modules. `Postulate.collect_exports` ignored visibility entirely, so a postulate declared `private` in one module was usable from another — unlike every other name-exporting declaration (`Union`, `Define`, `RecFun`, …) and the analogous module-private `lemma`. In the issue's repro this even let an importer "prove" `true = false` from another module's private axiom.

## Fix

Add the standard private-visibility guard to `Postulate.collect_exports`, matching `ProcDecl`/`Define`/`Theorem`:

```python
def collect_exports(self, export_env, importing_module):
    if self.visibility != 'private' or importing_module == get_current_module():
      export_env[base_name(self.name)] = [self.name]
```

A `private postulate` is now hidden from importers (importer gets the ordinary `undefined proof variable` error), while remaining usable within its own module. Public postulates are unaffected.

## Verification

Repro from the issue now errors instead of validating:

```
/tmp/pt/useit.pf:2.41-2.52: undefined proof variable secret_fact
```

Confirmed all three cases:
- private postulate from another module → hidden (error) ✓
- public postulate from another module → visible (valid) ✓
- private postulate within the same module → visible (valid) ✓

## Tests

- Added `private postulate cant_import_postulate` to `test/test-imports/Private.pf` (alongside the existing private `lemma`/`union`).
- New `test/should-error/import_postulate.pf` + `.err` fixture mirroring `import_lemma.pf`, asserting the private postulate is not importable.
- Full `python3 test-deduce.py` (both parsers, should-validate/error/warn, parser equivalence) passes.

Closes #1065

Resume this session on ginger: `~/deduce-runner/resume.sh 3dc6eb8b-b61e-45bd-84cc-46a3f96bbc86 routine/20260718-040202`

### DIFF
--- a/abstract_syntax/declarations.py
+++ b/abstract_syntax/declarations.py
@@ -562,7 +562,8 @@ class Postulate(Declaration):
                      visibility=self.visibility)
 
   def collect_exports(self, export_env: UniquifyEnv, importing_module: str) -> None:
-    export_env[base_name(self.name)] = [self.name]
+    if self.visibility != 'private' or importing_module == get_current_module():
+      export_env[base_name(self.name)] = [self.name]
 
 @dataclass
 class Theorem(Declaration):
--- a/abstract_syntax/theorems.py
+++ b/abstract_syntax/theorems.py
@@ -57,7 +57,8 @@ def collect_public(s: Statement, to_print: list[TheoremFilePrinting]) -> None:
       if not s.isLemma:
         to_print.append(s)
     elif isinstance(s, Postulate):
-      to_print.append(s)
+      if s.visibility != 'private':
+        to_print.append(s)
     elif isinstance(s, Auto):
       if not _auto_target_is_private(s.name):
         to_print.append(s)
--- a/test/should-error/import_postulate.pf
+++ b/test/should-error/import_postulate.pf
@@ -1,0 +1,6 @@
+import Private
+
+theorem T: true
+proof
+  cant_import_postulate
+end
--- a/test/should-error/import_postulate.pf.err
+++ b/test/should-error/import_postulate.pf.err
@@ -1,0 +1,1 @@
+./test/should-error/import_postulate.pf:5.3-5.24: undefined proof variable cant_import_postulate
--- a/test/test-imports/Private.pf
+++ b/test/test-imports/Private.pf
@@ -7,3 +7,5 @@ lemma cant_import_lemma: true
 proof
   .
 end
+
+private postulate cant_import_postulate: true


jsiek:
Good catch — addressed in e7b750b. `collect_public` in `abstract_syntax/theorems.py` was appending every `Postulate` unconditionally, so a `private postulate` still leaked into the generated `.thm` summary (a second public export surface). Added the same private-visibility filter the `Theorem` branch uses (and that `_collect_private_proof_names` already relies on). Verified: a module with `private postulate secret: ...` no longer emits `secret` to its `.thm`, while public postulates are still exported. Full `test-deduce.py` passes.

PR #1071, opened 2026-07-18
Add ASCII synonyms for the ≈, ≲, and ⊝ operators (closes #1069)

### LINKED ISSUE #1069: RD parser lists `~~`/`<~` as ASCII synonyms for `≈`/`≲` but neither lexer accepts `~` (dead entries)
**Is this intentional? If so, close.**

### One-sentence bug

The recursive-descent parser lists `~~` and `<~` as ASCII synonyms for `≈` and `≲` (in `compare_operators` and the `to_unicode` map), but neither parser's lexer can produce a `~` token, so those two synonyms are unreachable dead entries.

### Where

`rec_desc_parser.py`:

```python
compare_operators = {'<', '>', '≤', '<=', '≥', '>=', '⊆', '(=', '∈', 'in', '≲', '<~', '≈', '~~'}
...
to_unicode = {..., '~~': '≈', '<~': '≲', ...}
```

Every other `to_unicode` entry (`<=`→`≤`, `.o.`→`∘`, `(=`→`⊆`, `in`→`∈`, `<=>`→`⇔`, …) is a working ASCII synonym that both parsers accept. `~~`/`<~` are the only ones that do not lex: the `OPERATOR` terminal in `Deduce.lark:6` (`/[\-\+∸*%=≠<>≤≥&∘∪∩⊆∈⨄⊝\^≲≈]/`) has no `~`, and the RD lexer rejects `~` the same way.

### Minimal repro

```sh
$ echo 'define t = f ~~ g' | python3 deduce.py /dev/stdin
# <stdin>:1.14-1.15: unexpected character '~'

$ echo 'define t = f <~ g' | python3 deduce.py /dev/stdin
# <stdin>:1.15-1.16: unexpected character '~'
```

The Unicode forms `f ≈ g` / `f ≲ g` parse fine under both parsers.

### Observed vs expected

- Observed: `~~` / `<~` are rejected at lex time by both parsers, so the `to_unicode`/operator-set entries are dead.
- Expected: either (a) they are genuine ASCII synonyms and should lex to `≈`/`≲` (add `~` to the lexer and a LALR grammar alternative), or (b) they are not intended and the two dead entries should be removed from `rec_desc_parser.py`. `≈`/`≲` have no ASCII synonym today, unlike every other operator, so (a) would restore consistency; (b) is the smaller change.

### Found via

Surfaced while fixing the `≈` RD/LALR precedence divergence (refs #473): an inline parse-equivalence probe `a ~~ b < c` failed to parse under *both* parsers, revealing the entries never lex. Not fixed in that PR to keep it scoped to the precedence alignment.

Refs #473.

jsiek:
## Summary

Every Unicode infix operator in Deduce already had an ASCII spelling — `<=`
for `≤`, `.o.` for `∘`, `(=` for `⊆`, `.+.` for `⨄`, and so on — **except**
the two approximate comparisons `≈`/`≲` and the circled-minus signed
difference `⊝`. This PR closes that gap so every Unicode operator has an ASCII
equivalent (issue #1069):

| Unicode | ASCII | operator | precedence |
|:-:|:-:|:--|:--|
| `≈` | `~~` | approx-equal | comparison |
| `≲` | `<~` | approx-less-equal | comparison |
| `⊝` | `~-` | signed difference | additive |

The recursive-descent parser already *listed* `~~`/`<~` in its operator
tables and `to_unicode` map, but neither lexer could produce them (the
`OPERATOR` terminal has no `~`), so they were unreachable dead entries. This
makes them real.

> Note: this replaces the original direction of this PR (which removed the
> two dead entries). Per maintainer request, the operators are now wired up
> as genuine ASCII synonyms instead.

## Changes

- **`Deduce.lark`** — add `~~`/`<~`/`~-` as leading-underscore *named*
  terminals. Named (not inline string literals) so they don't renumber the
  `__ANON` series that `keywords.py` mirrors; leading-underscore so lark
  filters them from the parse tree exactly as it filters the inline operator
  literals — otherwise the generic infix transformer in `parser.py` (which
  maps over *every* child) would pick the operator token up as a bogus
  operand. Adds the infix rules and the `operator ~~`/`<~`/`~-` declaration
  forms.
- **`parser.py`** — map the new `identifier_*` declaration aliases to the
  Unicode operator names.
- **`rec_desc_parser.py`** — add `~-` to `add_operators` and `to_unicode`
  (`~~`/`<~` were already present from #1070).
- **`gh_pages/scripts/keywords.py`** — register the three named terminals.
- **`gh_pages/js/codeKeywords.js`** — add the three to the syntax-highlight
  operator list (placed after their Unicode counterparts, matching the
  existing `.o.`/`.+.` convention and regex-priority ordering).
- **`gh_pages/doc/Reference.md`** — note the ASCII spellings in the operator
  precedence table.

## Testing

- `python3 test-deduce.py` — full default suite passes (lib, should-validate ×
  both parsers, should-error, should-warn, parser equivalence, round-trip).
- `python3 gh_pages/scripts/keywords.py` and `reference_grammar.py` (the CI
  static token/grammar checks) pass; the `__ANON` numbering is unchanged.
- New `test/should-validate/ascii_operator_synonyms.pf` exercises all three
  operators as `operator` declarations and as infix uses (validates under both
  parsers).
- New `test-deduce.py` equivalence guards assert each ASCII form parses to the
  *same* AST as its Unicode counterpart, under both RD and LALR.

Since `~` never lexed before, no existing file can use these spellings for
anything else, so the change is purely additive.

Closes #1069.

Resume this session on ginger: `~/deduce-runner/resume.sh 623e1580-97c8-4252-b439-92ba41a42b2b routine/20260718-070201`


jsiek:
Redirected this PR per maintainer request: instead of *removing* the dead `~~`/`<~` entries, it now wires up `~~` (≈), `<~` (≲), and `~-` (⊝) as real ASCII synonyms, so every Unicode operator in Deduce has an ASCII spelling. The branch was force-pushed (based on current `main`), which also clears the earlier merge conflict with #1070.

Issue #813, opened 2026-06-03
Reduce code duplication across files


jsiek:
Look for similar functionality across different files, and when you find it create shared functions that can be reused in those files. When the shared functions already have an appropriate file that they can live in, go ahead and put them there, if not create a new file with a new charter for the appropriate contents of that file.

Pull requests should be gated on whether the PR actually reduces the number of lines of code, not counting comment strings.

## Current status

Focused slices have extracted several duplicated checker helpers:

- PR #959 extracts the repeated function-type alpha-renaming logic in `checker_pipeline.py`.
- Draft PR #975 extracts the duplicated switch-pattern constructor matching heuristic shared by term checking and proof checking.

The umbrella remains open for additional duplication-reduction opportunities across the codebase.

jsiek:
Claude is taking this issue in progress for a focused slice: extracting the duplicated `Constructor` parameter-instantiation logic. Three call sites in `checker_types.py` and `checker_proofs.py` build the same `typarams → type_args` substitution and apply it to `constr.parameters`. Will lift that into a `Constructor.instantiate_parameters(typarams, type_args)` method on the `Constructor` dataclass in `abstract_syntax/declarations.py` and update the call sites.

[Claude session](claude://resume?session=7027ee84-2e5f-4c1d-ad4a-1050153c8228) — fallback UUID: `7027ee84-2e5f-4c1d-ad4a-1050153c8228`

jsiek:
[#938](https://github.com/jsiek/deduce/pull/938) merged: lifted the duplicated `typarams → type_args` substitution over a union constructor's parameter types into a `Constructor.instantiate_parameters` method, replacing three near-identical blocks in `checker_types.check_constructor_pattern`, `checker_proofs._check_proof_of_induction`, and `checker_proofs._check_proof_of_switch`.

Umbrella stays open — many other duplications remain across the codebase.

jsiek:
Claude is taking another focused slice on this umbrella: lifting the duplicated `FunctionType` type-argument instantiation pattern. The same six-line block (build `typarams → tyargs` sub, substitute in `param_types` / `return_type`, build flat `FunctionType` with empty type_params) appears in both `checker_types.type_check_term_inst` and `checker_types.type_check_term_inst_var`. Will add a `FunctionType.instantiate(tyargs)` method on the `FunctionType` dataclass in `abstract_syntax/terms.py` (mirroring the `Constructor.instantiate_parameters` helper from #938) and update the two call sites.

[...]

PR #1075, opened 2026-07-18
Collapse duplicated VarRef method bodies into the shared base (refs #813)

jsiek:
## Summary

Removes duplication in the `VarRef` hierarchy (`abstract_syntax/terms.py`), part of the #813 code-dedup umbrella.

The three `VarRef` subclasses each carried near-identical method bodies:

- `Var.substitute` and `ResolvedVar.substitute` were byte-for-byte identical except the reconstructed class name; `OverloadedVar.substitute` differed only in reading `resolved_names[0]` instead of `.name` (which is exactly what its `name` property / `get_name()` return).
- `OverloadedVar.reduce` and `ResolvedVar.reduce` were identical modulo the same `resolved_names[0]` vs `.name` spelling.
- All three `__str__` methods differed only in the `--unique-names` brace form.
- All three `copy` methods differed only in the reconstructed class.

This PR:

- moves `copy`, `substitute`, `__str__`, and the post-uniquify `reduce` onto the `VarRef` base class, driving them through `get_name()` plus two small per-subclass hooks:
  - `_rebuild(typeof)` — reconstruct this reference carrying a (possibly new) type annotation;
  - `_unique_names_str()` — the `name{...}` form shown under `--unique-names`;
- keeps `Var.reduce` as its pre-uniquify no-op override;
- leaves each `__eq__` per-subclass on purpose — the pre-/post-uniquify comparison asymmetry (and the cross-boundary `OverloadedVar`↔`ResolvedVar` matching) is deliberate and documented in the existing comments;
- refreshes the `VarRef` docstring and the stale `subst_typeof` comment in `core.py` that referenced the now-removed per-class `substitute` methods.

Net **−33 lines** (94 insertions, 127 deletions).

## Behavior preserved

No behavior change. Each subclass's `get_name()` already returned the exact name the old bodies read (`OverloadedVar.name` is a property equal to `resolved_names[0]`), and `_rebuild`/`_unique_names_str` reproduce the old per-class reconstruction and brace-form output verbatim.

## Verification

- `python3 test-deduce.py` — all categories green (lib + should-validate ×2 parsers + should-error + should-warn + prelude + parser equivalence), 144s.
- `python3 deduce.py --unique-names <file>` spot-check (exercises `_unique_names_str`); evaluation traces in the should-validate corpus exercise the shared `reduce`.
- ruff/mypy could not run in this container (no `python3.13`); relying on CI static checks. The moved method signatures match the originals (including the existing `# type: ignore[override]` on `reduce`).

Resume this session on ginger: `~/deduce-runner/resume.sh 857e5e0b-a698-455d-a534-15a0390dc079 routine/20260718-100202`

### DIFF
--- a/abstract_syntax/core.py
+++ b/abstract_syntax/core.py
@@ -189,9 +189,8 @@ class Term(AST):
 
   def subst_typeof(self, sub: Mapping[str, Term | Type | RecFun | GenRecFun]) -> Optional[Type]:
     # Apply `sub` to the cached `typeof` annotation. Kept as a named
-    # helper because a handful of bespoke `substitute` methods (Var,
-    # OverloadedVar, ResolvedVar) call it directly to short-circuit
-    # when the annotation didn't actually change.
+    # helper because the shared `VarRef.substitute` calls it directly to
+    # short-circuit when the annotation didn't actually change.
     return self.typeof.substitute(sub) if self.typeof is not None else None
 
   def substitute(self, sub: Mapping[str, Term | Type | RecFun | GenRecFun]) -> Term | RecFun | GenRecFun:
--- a/abstract_syntax/terms.py
+++ b/abstract_syntax/terms.py
@@ -454,11 +454,18 @@ class TAnnote(Term):
 @dataclass
 class VarRef(Term):
   # Abstract base for variable references. Concrete subclasses are
-  # `Var` (post-parse, holds a source identifier) and `OverloadedVar`
-  # (post-uniquify, holds uniquified candidate names). Code that wants
-  # to operate on either variant should `isinstance(node, VarRef)` and
-  # call `get_name()` rather than reach for a `name` field — the two
+  # `Var` (post-parse, holds a source identifier), `OverloadedVar`
+  # (post-uniquify, holds uniquified candidate names), and `ResolvedVar`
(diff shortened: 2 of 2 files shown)


jsiek:
Rebased onto latest `main` to resolve a merge conflict. The conflict was with #1062's UInt-zero fix (merged as #1063/#1064), which had added an `isBZero(self) → '0'` branch to `ResolvedVar.__str__` — the method this PR removes by consolidating into the base `VarRef.__str__`.

Resolution: added a small `_special_str()` hook to the shared `VarRef.__str__` (default `None`), overridden only by `ResolvedVar` to return `'0'` for `bzero`. This preserves the exact prior behavior, **including the branch ordering** (the `bzero → '0'` case still wins over the `--unique-names` brace form) and the fact that only the *resolved* form is special-cased — an `OverloadedVar` for `bzero` still prints `bzero`, exactly as before this PR.

Verified after rebase: `python3 test-deduce.py` all green (170s); `print (0 : UInt)` prints `0` with and without `--unique-names`.

Note (not changed here): `isBZero` matches both `OverloadedVar` and `ResolvedVar`, but `__str__` only special-cases the resolved form, so an unresolved `OverloadedVar` bzero prints `bzero` rather than `0`. That asymmetry predates this PR and is preserved verbatim; flagging in case it was an oversight in #1062.

Issue #720, opened 2026-05-21
Stdlib Int: Euclidean division, modulo, divisibility, gcd, lcm


jsiek:
Spun out from #660 (now closed) — Slice 3, the largest remaining practical completeness gap in the Int stdlib.

## Scope

Deduce has `Int` division (`lib/IntDefs.pf`), but no `Int` modulo operator, `divides` predicate, `gcd`, `lcm`, or the rich `UIntDiv.pf`-style theorem set.

## Candidates

- **Decide and document the rounding convention** (Euclidean / truncated / floored) before implementation. Lean uses Euclidean `ediv`/`emod`; Coq `ZArith`, Isabelle/HOL, and Agda each provide useful comparison points.
- Define `%` / `mod` for `Int`, with the documented rounding convention.
- State and prove `div_mod`: quotient/remainder equation plus remainder bounds.
- `mod_small`, `mod_self`, `mod_one`, `div_one`, `div_zero` policy, and sign / `abs` interaction.
- `divides` over `Int`; closure under `+`, `-`, `*`; antisymmetry up to sign or absolute value as appropriate.
- `gcd` / `lcm` over `Int`, probably defined through `abs` / `UInt.gcd`, with divisibility / greatest properties and Bezout-style follow-up potential.

## External references

- Lean Euclidean `Int.ediv` / `Int.emod` convention: https://leanprover-community.github.io/mathlib4_docs/Init/Data/Int/DivMod/Basic.html
- Lean/mathlib integer gcd/lcm and Bezout support: https://leanprover-community.github.io/mathlib4_docs/Mathlib/Data/Int/GCD.html
- Isabelle/HOL integer division/modulo implementation and laws: https://isabelle.in.tum.de/repos/isabelle/file/b1d955791529/src/HOL/Divides.thy
- Isabelle/HOL gcd/lcm on `int`: https://isabelle.in.tum.de/website-Isabelle2022/dist/library/HOL/HOL/GCD.html
- Coq integer arithmetic / divisibility / gcd: https://coq.inria.fr/library/Coq.ZArith.BinInt.html and https://coq.inria.fr/library/Coq.ZArith.Znumtheory.html
- Agda integer division/modulo: https://agda.github.io/agda-stdlib/v2.1/Data.Integer.DivMod.html
- Agda integer gcd: https://agda.github.io/agda-stdlib/v1.7.3/Data.Integer.GCD.html

## Single-slice instruction

[...]

PR #1080, opened 2026-07-18
lib: bridge Int divisibility to the truncated remainder (refs #720)

jsiek:
## Summary

Adds the divisibility↔remainder bridge to `lib/IntDivides.pf`, one of the
remaining sign/`abs`-interaction laws in the `Int` div/mod/gcd story tracked by
#720:

- **`int_divides_iff_mod_zero: if not (m = +0) then (divides(m, n) ⇔ n % m = +0)`**
  — for a nonzero divisor, `m` divides `n` exactly when the truncated remainder
  `n % m` vanishes. This is the foundational bridge connecting the `divides`
  predicate (`lib/IntDivides.pf`) to the `%` operator (`lib/IntMod.pf`).

### Proof shape

- **Forward** (`divides(m, n) → n % m = +0`): from `m * k = n`, show
  `abs(n % m) = 0` by routing through `int_abs_mod`, `int_abs_mult`, and
  `uint_mult_mod_left_zero` (using `0 < abs(m)` from `m ≠ +0`), then conclude via
  `int_abs_eq_zero_implies_zero`.
- **Backward** (`n % m = +0 → divides(m, n)`): substitute the remainder into the
  `int_div_mod` identity `(n / m) * m + (n % m) = n` to get `(n / m) * m = n`, and
  choose `n / m` as the witness.

`IntDivides.pf` now imports `IntMod` (no import cycle: `IntMod` depends only on
`{UInt, Base, IntDefs, IntAddSub, IntMult, IntLess, IntAbs, IntDiv}`, none of
which depend on `IntDivides`).

## Theorem-summary generation fix (addresses Codex review)

The new private `import IntMod` inside `IntDivides.pf` exposed a latent bug in
`abstract_syntax/theorems.py`. `collect_public` added a module to
`collected_imports` even for a **private** import, so when a file publicly
re-exported a module that an earlier public import had already reached
privately, the later public import was skipped and that module's public
theorems were dropped from the generated `.thm` summary.

Concretely, `lib/Int.pf` does `public import IntDivides` (line 24) before
`public import IntMod` (line 25); the private `import IntMod` inside
`IntDivides` marked `IntMod` collected, so `Int.thm` lost the entire public `%`
/ IntMod surface.

Fix: only mark a module collected when we actually recurse into it (i.e. on a
public import). Verified across the whole `lib/` tree that the change is purely
additive (0 lines removed from any `.thm`):

- `Int.thm` regains its IntMod public surface (`%`, div/mod laws, Even/Odd).
- `Nat.thm` regains `EvenNat`/`OddNat` — a **pre-existing** instance of the same
  drop (via `NatDiv`'s private `import NatEvenOdd`, before `Nat.pf`'s
  `public import NatEvenOdd`), fixed here as a bonus.

## Testing

- `python3 deduce.py lib/IntDivides.pf` and `--lalr` → valid (both parsers)
- `python3 test-deduce.py` (full default suite: lib × 2 parsers, should-validate,
  should-error, should-warn, parser equivalence) → all pass
- `python3 gh_pages/scripts/keywords.py` and `reference_grammar.py` → pass
- Regenerated every `lib/*.thm` before/after the `theorems.py` change and
  confirmed only `Int.thm` and `Nat.thm` differ, both purely additive.

## Scope note

`#720` remains open after this PR — Bezout-style identities and other remaining
laws are not covered here, so this uses `Refs`, not `Closes`.

Refs #720.

Resume this session on ginger: `~/deduce-runner/resume.sh 600c1d6b-7d19-4b98-93db-02c9c3bd946e routine/20260718-140202`

### DIFF
--- a/abstract_syntax/theorems.py
+++ b/abstract_syntax/theorems.py
@@ -65,8 +65,12 @@ def collect_public(s: Statement, to_print: list[TheoremFilePrinting]) -> None:
     elif isinstance(s, Import):
       if s.name in collected_imports:
           return
-      collected_imports.add(s.name)
+      # Only a public import contributes (and dedups) the module's public
+      # theorems. A private import must not mark the module collected, or a
+      # later public re-export of the same module would be skipped and its
+      # public theorems silently dropped from the .thm summary.
       if s.ast != None and s.visibility == 'public':
+        collected_imports.add(s.name)
         for stmt in s.ast:
             collect_public(stmt, to_print)
     elif isinstance(s, Declaration) and not s.visibility == 'private':
--- a/lib/IntDivides.pf
+++ b/lib/IntDivides.pf
@@ -7,6 +7,7 @@ import IntAddSub
 import IntMult
 import IntLess
 import IntAbs
+import IntMod
 
 /*
   Integer divisibility, gcd, and lcm.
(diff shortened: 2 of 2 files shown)


jsiek:
Thanks — good catch. Fixed at the root in `abstract_syntax/theorems.py`: `collect_public` no longer marks a module `collected` on a *private* import, so a later public re-export is still recursed into. This restores the full IntMod public surface to `Int.thm`, and as a bonus restores `EvenNat`/`OddNat` to `Nat.thm` (a pre-existing instance of the same drop via `NatDiv`'s private `import NatEvenOdd`). Verified across the whole `lib/` tree that the `.thm` regeneration diff is purely additive (0 lines removed).

Issue #931, opened 2026-06-10
Pretty-printer round-trip coverage umbrella


jsiek:
## Current status

This is now the umbrella issue for pretty-printer/parser round-trip coverage, not a live snapshot of the original 110 failing `should-validate` files.

The original report came from a full local audit during #930 prep. At that time, many `test/should-validate/*.pf` files either failed to reparse after pretty-printing or reparsed to a different canonical AST. Since then, the focused PR series has addressed the original `should-validate` parse-fail and AST-drift clusters.

As of June 13, 2026:

- The original eight bug categories listed below have fixes in place.
- The `test/should-validate/*.pf` round-trip parse-fail set was driven to 0 in the follow-up PR series.
- The remaining purpose of this issue is broader round-trip coverage, especially the `lib/` corpus and any future full-corpus audit results.
- New fixes should continue adding representative files to `PARSER_ROUND_TRIP_FILES`.

## Relation to #980

#980 changes the CI/runtime shape of `test-deduce.py --equiv`; it does not change the correctness goal of this issue.

After #980, `--equiv` is intentionally a curated CI corpus so the GitHub Actions parser-equivalence job stays fast. It still uses `PARSER_ROUND_TRIP_FILES` as the main round-trip coverage set, but it no longer means "sweep every checked-in `.pf` file."

Implications for this issue:

- Keep expanding `PARSER_ROUND_TRIP_FILES` with representative files as round-trip bugs are fixed.
- Treat full-corpus round-trip sweeps as local/manual audits, or add a separate opt-in harness mode if we want that audit automated.
- Do not assume default CI covers every `lib/`, `should-validate`, or `should-error` file for parser-equivalence drift after #980.

## Fixed original categories

The original issue enumerated these bug families. They are retained here as historical context, but they are no longer the active remaining set.

- [x] Numeric literal sugar was lost during pretty-printing. Fixed by #935.
- [x] `private` visibility was silently dropped. Fixed by #935 and completed for theorem/postulate declarations by #944.
- [x] `define name : type = fun ...` was rewritten to `fun name(...) ...`. Fixed by #935.
- [x] `conclude ... by .` and related one-token reasons were not parse-preserving. Fixed in the #935/#937 series.
- [x] `switch case <pat> assume <name> { ... }` dropped the `assume` binding. Fixed by #935.
- [x] `simplify with q` dropped `with`. Fixed by #935.
[...]

PR #1090, opened 2026-07-24
Add ≠ to infix_precedence so `operator ≠` round-trips (refs #931)

jsiek:
## Summary

A fresh full-corpus `--equiv-full` round-trip sweep against `main` surfaced 8
failures across two `should-error` files that reference an operator by name in
term position:

```
define d = operator ≠
define d = operator /=   // /= normalizes to ≠ at parse time
```

Both parse to `Var('≠')`, but the pretty-printer emitted a bare `≠`, which
neither parser accepts as a term — so the pretty-printed source failed to
reparse under both RD and LALR.

## Root cause

`VarRef.__str__` only prepends the `operator ` keyword when
`is_var_operator` → `is_operator_name(name)` is true, and
`is_operator_name` is defined purely in terms of the `infix_precedence` /
`prefix_precedence` tables in `abstract_syntax/core.py`. `≠` was **absent**
from `infix_precedence` (even though `=`, `<`, `≤`, `≥`, `>`, `∈`, `⊆`, … all
appear there), so `is_operator_name('≠')` was `False` and the keyword was
dropped.

## Fix

Add `'≠': 1` to `infix_precedence` — `≠` is a precedence-1 comparison operator,
identical in kind to `=`. This is safe because:

- Neither parser imports `infix_precedence`; only the pretty-printer consults it
  (grep-verified). So parsing behavior is unchanged.
- `≠` is always desugared to `if (x = y) then false` by both parsers, so there
  are no `≠`-headed `Call` nodes for `precedence()` to newly parenthesize; the
  existing `IfThen(_, _, _, Bool false)` case still handles the `≠` sugar.
- The undefined-variable error for `operator ≠` now reads
  `undefined variable operator ≠`, matching how `operator =` and every other
  operator name already print (`operator =` already yielded
  `undefined variable operator =`). `≠` was the lone inconsistency.

## Changes

- `abstract_syntax/core.py`: add `'≠': 1` to `infix_precedence`.
- `test-deduce.py`: add both `should-error` files to `PARSER_ROUND_TRIP_FILES`.
- Regenerate the two `.err` fixtures (message now names `operator ≠`; the
  regression-guard intent — "the error must name `≠`" — still holds).

## Verification

- `python3 test-deduce.py --equiv` — pass (curated equiv + round-trip).
- `python3 test-deduce.py --equiv-full` — **533/533 files round-trip; 0
  failures** (was 8). No new drift introduced by the table change.
- `python3 test-deduce.py --errors` — pass.
- `python3 test-deduce.py --passable --warns` — pass (both parsers).
- `keywords.py` and `reference_grammar.py` consistency checks — pass.

(`ruff`/`mypy` require `python3.13`, which is not present in this container; the
change is a one-key dict addition plus list/fixture edits with no type surface.)

Refs #931, #473.

Resume this session on ginger: `~/deduce-runner/resume.sh 577f6331-0541-4b22-8dd1-d41e0eedad7b routine/20260724-050201`


jsiek:
Closing as a duplicate of #1086, which predates this PR (routine run 00:02 vs 05:02 today) and fixes the identical `operator ≠` round-trip defect with the same root-cause analysis.

#1086's approach is also cleaner: it introduces a dedicated `sugar_operators = {'≠'}` set rather than adding `≠` to `infix_precedence` (this PR's approach). Since `≠` is pure sugar — `x ≠ y` always desugars to `not (x = y)`, so no `Call`/term node is ever headed by it — it doesn't semantically belong in the precedence table alongside real infix operators. #1086 also folds `complete_name` onto `is_operator_name` to dedup their identical conditions.

Deferring to #1086. No further action here.

PR #1091, opened 2026-07-24
Dedup: share header type-params `<...>` rendering across decl printers (refs #813)

jsiek:
## Summary

Small duplication-reduction slice for the #813 umbrella.

`RecFun`, `GenRecFun`, `ViewRecFun`, and `ViewDecl` each inlined a
byte-identical two-line expression to render the optional
`<t1,t2,...>` type-parameter suffix in their `pretty_print` headers:

```python
('<' + ','.join([name2str(t) for t in self.type_params]) + '>'
 if len(self.type_params) > 0 else '')
```

This PR extracts that into a shared `angle_type_params` helper in
`abstract_syntax/declarations.py` and points all four call sites at it.

## Why this is safe

Pure refactor — the four sites were textually identical (same `name2str`
render, same `,` separator, same `len(...) > 0` guard), so the extracted
helper reproduces each one exactly. It is deliberately scoped to the
`name2str`/`,` form only; the `base_name`- and raw-rendered type-param
lists elsewhere in the module render differently and are intentionally
left alone.

## Verification

- `python3 test-deduce.py --equiv` — pass (round-trip exercises the
  `recfun` / `viewrec` / `recursive` declaration printers via
  `recfun_roundtrip.pf`, `uint_viewrec.pf`, `fib.pf`, and the `lib/`
  modules).
- `python3 test-deduce.py --passable` — pass (both parsers).
- Spot-checked that `replicate<T>(...)`, `fib` (no params), and
  `count_down` (no params) headers render identically to before.

(`ruff`/`mypy` require `python3.13`, which is absent in this container; the
change is a helper extraction with no new type surface.)

Refs #813.

Resume this session on ginger: `~/deduce-runner/resume.sh 577f6331-0541-4b22-8dd1-d41e0eedad7b routine/20260724-050201`

### DIFF
--- a/abstract_syntax/declarations.py
+++ b/abstract_syntax/declarations.py
@@ -1212,8 +1212,7 @@ class RecFun(Declaration):
 
   def pretty_print(self, indent: int, afterNewline: bool = False) -> str:
     header = complete_name(self.name) \
-        + ('<' + ','.join([name2str(t) for t in self.type_params]) + '>' \
-           if len(self.type_params) > 0 else '') \
+        + angle_type_params(self.type_params) \
       + '(' + ','.join([str(ty) for ty in self.params]) + ')' \
       + ' -> ' + str(self.returns)
     ret = self.visibility_prefix() + 'recursive ' + header + '{\n' \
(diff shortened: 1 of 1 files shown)


jsiek:
This refactor increases total LOC instead of reducing it. The reason is that the duplicated code is only a single line of code, too small. The factoring into a separate function just makes the code harder to read.

PR #1092, opened 2026-07-24
Machine-check the Reference.md operator-precedence table against Deduce.lark (refs #473)

jsiek:
## Summary

Advances issue #473 goal B ("a check that Reference.md grammar excerpts agree with the LALR grammar") by machine-checking the **Operator Precedence** table in `Reference.md` against `Deduce.lark`.

That table documents the term grammar's precedence levels — each row naming a `Deduce.lark` rule and the operators it introduces — but nothing checked it against the grammar. It was correct at the time of this PR, yet could silently drift whenever an operator is added, moved between levels, or renamed in `Deduce.lark`. This closes that gap the same way the existing `deduce-grammar` blocks pin the per-operator syntax snippets.

## Changes

Extend `gh_pages/scripts/reference_grammar.py` with a `check_precedence_table` pass (run from `main`, so it fires everywhere the script already runs — `make tests-tokens`, `static.yml`, `test_deduce.yml`). Against `Deduce.lark` it verifies:

- **rule existence** — every rule named in the table (`atomic_term` … `iff_term`) is a real rule;
- **fall-through chain** — each looser level reaches the next tighter level *transitively* through the grammar's `| <rule>` fall-through alternatives, so the table's intentional omission of the experimental `sep_term`/`pointsto_term` levels between `logical_term` and `equality_term` is tolerated; and
- **operator membership** — every operator listed at a level, including the `(also …)` ASCII aliases (resolved through single-literal terminals such as `_APPROXEQ`, `_APPROXLE`, `_OMINUS`), is actually an operator of that rule in the grammar.

Level 0 (`atomic_term`) lists descriptive primary/prefix *forms* (`f(...)`, `@f<T>`, `not P`, `-n`, …) rather than plain operators, so only its rule name and its role as the chain's tight end are checked.

No changes to the table itself were needed — it already agrees with the grammar (47 precedence entries checked).

## Testing

- `python3 gh_pages/scripts/reference_grammar.py` → `Checked 177 reference grammar rule(s) and 47 operator-precedence entr(y/ies) against Deduce.lark.`
- Verified the new check actually catches drift by temporarily mutating the real files (each reverted afterward):
  - a bogus operator at a level → `operator '^' is documented at level 2 (multiplicative_term) but is not an operator of multiplicative_term in Deduce.lark`
  - a bad rule name → `rule 'bogus_term' is not in Deduce.lark`
  - a broken fall-through → `level 1 (exponent_term) does not fall through to level 0 (atomic_term) in Deduce.lark`
  - all with accurate `Reference.md` line numbers.
- `python3 -m ruff check .` and `python3 -m mypy .` — clean (54 source files).
- `python3 gh_pages/scripts/keywords.py` — clean.

## Note

This is a partial contribution to the umbrella issue #473 (the operator-precedence table item of goal B); other goal-A/B items remain, so this uses `Refs`, not a closing keyword.

Resume this session on ginger: `~/deduce-runner/resume.sh 840d3c34-7fe9-40c5-8a9b-896f86cee444 routine/20260724-060202`

### DIFF
--- a/gh_pages/doc/Reference.md
+++ b/gh_pages/doc/Reference.md
@@ -1904,7 +1904,7 @@ tighter rule listed above it.
 |     0 | `atomic_term`          | literals, identifiers, parentheses, `f(...)`, `a[i]`, `@f<T>`, `not P`, `-n`       | primary and prefix forms                                           |
 |     1 | `exponent_term`        | `^`                                                                                | exponent                                                           |
 |     2 | `multiplicative_term`  | `*`  `/`  `%`  `∘` (also `.o.`)                                                    | multiply, divide, modulo, function composition                     |
-|     3 | `additive_term`        | `+`  `-`  `∸` (also `.-.`)  `⊝` (also `~-`)  `++`  `∪`  `∩`  `⨄` (also `.+.`)      | add, subtract, monus, list append, set/multiset union/intersection |
+|     3 | `additive_term`        | `+`  `-`  `∸` (also `.-.`)  `⊝` (also `~-`)  `++`  `∪`  `∩` (also `&`)  `⨄` (also `.+.`)  | add, subtract, monus, list append, set/multiset union/intersection |
 |     4 | `comparison_term`      | `<`  `>`  `≤` (also `<=`)  `≥` (also `>=`)  `⊆` (also `(=`)  `∈` (also `in`)  `≲` (also `<~`)  `≈` (also `~~`) | comparisons, subset, membership                                |
 |     5 | `equality_term`        | `=`  `≠` (also `/=`)                                                               | equality, inequality                                               |
 |     6 | `logical_term`         | `and`  `or`  `:`                                                                   | logical conjunction, disjunction, type annotation                  |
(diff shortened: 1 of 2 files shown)


jsiek:
Good catch — addressed in f62c42f. The operator membership check is now **bidirectional** for levels 1+: in addition to `documented ⊆ grammar`, each grammar operator of a level's rule must also be documented at that level, so a new operator added to `Deduce.lark` but not to the table now fails CI (the primary drift this check exists to catch).

While making it symmetric I also escaped away one of the two exceptions: `&` (∩) renders correctly in a code span, so it is now listed in the table directly rather than carved out. Only `|` (∪) remains an exception, because it is the Markdown column separator and `\|` renders with a stray backslash under the site's `markdown` renderer — it stays documented in prose below the table.

Verified the reverse check catches drift: temporarily adding `multiplicative_term "⊗" exponent_term` to `Deduce.lark` produces `operator '⊗' is a level-2 (multiplicative_term) operator in Deduce.lark but is missing from the precedence table`.

Issue #1135, opened 2026-07-27
RD/LALR disagree on visibility-prefixed declaration .location start column


jsiek:
## Bug

For a **visibility-prefixed declaration** (`public`/`private`/`opaque`), the recursive-descent and LALR parsers disagree on the *start* of the declaration's `.location` span: RD anchors it at the declaration keyword, LALR at the visibility modifier.

## Minimal repro

`/tmp/loc.pf`:
```
public proc foo()
  ensures true
{
}
```

```sh
python3 deduce.py --recursive-descent --experimental-imperative --no-stdlib /tmp/loc.pf
python3 deduce.py --lalr            --experimental-imperative --no-stdlib /tmp/loc.pf
```

## Observed vs expected

The new Phase 1m "accepted but not verified" warning (issue #1108) is source-located at `decl.location`, so the divergence is now visible:

- RD:   `/tmp/loc.pf:1.8-4.2: warning: proc 'foo' is accepted but not verified ...`  (column 8 = the `proc` keyword)
- LALR: `/tmp/loc.pf:1.1-4.2: warning: proc 'foo' is accepted but not verified ...`  (column 1 = the `public` modifier)

Expected: both parsers produce the same `.location` for the same declaration (the AST is supposed to be parser-independent). A declaration with no visibility keyword (e.g. the `default_visibility` observer in `test/should-warn/observer_declarations.pf`) matches at column 1 under both parsers, which pins the difference to the visibility prefix.

## Impact / why CI stays green

- The `--warns` and `--errors` harnesses are RD-only, so `.pf.warn` fixtures generated under RD keep matching.
- The `--equiv` AST comparison ignores `location`, so the divergence does not fail parser-equivalence today.

So this is latent: any feature that surfaces `decl.location` for a visibility-prefixed declaration (diagnostics, LSP ranges) will diverge between parsers.

## Where surfaced

Surfaced while implementing #1108 (Phase 1m warnings); the warning reads `ProcDecl`/`ObserverDecl`/`ResourceDecl` `.location` in `checker_pipeline.check_proofs`. The divergence is in how `rec_desc_parser.py` vs `parser.py` set the declaration node's `location`/`Meta` when a visibility keyword precedes the declaration keyword.

Refs #473 (dual-parser equivalence umbrella), #854 (imperative layer).

jsiek:
[...]

PR #1136, opened 2026-07-27
Warn when imperative declarations are accepted without verification (closes #1108)

### LINKED ISSUE #1108: Phase 1m: Warn when imperative declarations are accepted without verification
Parent: #854

Phase: 1 finishing work and transition guard for Phase 2.

## Goal

Stop silently presenting unchecked imperative bodies and specifications as fully valid while preserving the Phase 1 import/export and tooling plumbing.

## Scope

- Emit one clear source-located warning for each `proc`, observer body, or resource body that is accepted without semantic verification.
- State that parsing and declaration plumbing succeeded but verification did not run.
- Avoid duplicate warnings when a module is imported more than once.
- Keep object declaration type checking unchanged.
- Provide a small helper that Phase 2 slices can retire per construct as verification support lands.
- Move or add fixtures under `test/should-warn` so warning text is regression-tested.

## Acceptance

- [ ] A deliberately false procedure such as `ensures result = true` with `return false` does not pass silently.
- [ ] Imperative import/export fixtures still work and warnings are deterministic.
- [ ] Both parser modes produce equivalent warning locations and text where practical.
- [ ] Existing pure Deduce files produce no new warnings.
- [ ] `python test-deduce.py --warns` and the default regression sweep pass.

## Dependencies

None. Land before the Phase 2 imperative harness so the harness has an explicit baseline.

## Out of scope

No verification, proof checking, or change to the meaning of any imperative construct.

jsiek:
Closes #1108

## Summary

Implements **Phase 1m** (issue #1108): stop silently presenting unchecked imperative declarations as fully valid, ahead of the Phase 2 verifier.

`check_proofs` now emits one source-located warning per `proc` / `observer` / `resource` declaration:

```
warning: proc 'always_false' is accepted but not verified -- parsing and declaration
plumbing succeeded, but no verifier has run on its body or specs yet (experimental
imperative layer, issue #854)
```

`object` declarations stay silent — they are genuinely type-checked in the earlier phases.

## Changes

- **`checker_pipeline.py`**
  - Add `warn_unverified_imperative(decl)`, a small helper the Phase 2 slices retire per construct (when procedure verification lands, drop the call from the `ProcDecl` case and verify there; observers/resources keep warning until their own verifier arrives). Split `ObjectDecl` out of the shared no-op `check_proofs` case so only the three unverified constructs warn.
  - Bypass the check_proofs cache for `ProcDecl`/`ObserverDecl`/`ResourceDecl`, exactly like `Print`/`Assert`. Their only effect in `check_proofs` is the warning, so with no warning sink installed (`check_file(collect_errors=False)`, e.g. the CLI) the cache-miss branch would otherwise cache them and a later re-check in a long-lived process (LSP/MCP) would silently drop the warning. *(Addresses the Codex review on this PR.)*
- **Fixtures** — move `proc_declarations.pf` and `observer_declarations.pf` from `should-validate` to `should-warn` (with generated `.warn` files), add `imperative_false_proc.pf` pinning the acceptance scenario (`ensures result = true` + `return false` must not pass silently).
- **`test-deduce.py`** — `_check_against_warn` now enables `--experimental-imperative` for enrolled files (it previously only did so for the `.err` path), and the round-trip / experimental-file path lists are updated to the new `should-warn` locations.

## Acceptance criteria (from #1108)

- [x] A deliberately false procedure does not pass silently — `imperative_false_proc.pf` + its `.warn`.
- [x] Imperative import/export fixtures still work; warnings are deterministic (top-level file is always freshly checked; imported modules warn once via the existing `checked_modules` guard). Verified a 3× re-check of one module in a single process warns every time.
- [x] Both parser modes produce equivalent warning locations and text. (The RD/LALR start-column divergence on visibility-prefixed declarations that this PR originally filed as #1135 has since been fixed on main by #1137; after rebasing, RD and LALR agree and the `.warn` fixtures were regenerated to match.)
- [x] Existing pure Deduce files produce no new warnings — the warned nodes only parse under `--experimental-imperative`.
- [x] `python3 test-deduce.py --warns` and the full default sweep pass.

## Testing

- `python3 test-deduce.py` (default: lib + should-validate ×2 parsers + should-error + should-warn + equiv) — all passed.
- `python3 test-deduce.py --equiv` — passed.
- `ruff check .` and `mypy .` — clean.
- Manual RD/LALR parity on the new fixtures — warning text and locations identical.
- Repro of the caching concern: checking one imperative module 3× in a single process with `collect_errors=False` warns on every run with the fix (warned only on run 1 without it).

## Notes

- Bug #1135 (RD/LALR visibility-prefixed declaration `.location` divergence), filed while implementing this PR, was fixed independently on main by #1137.

Resume this session on ginger: `~/deduce-runner/resume.sh e4d5d57d-ca7e-44d9-9d2a-b7dda56e9add routine/20260727-150202`


jsiek:
Addressed the caching finding in 9e56a81: `ProcDecl`/`ObserverDecl`/`ResourceDecl` now bypass the check_proofs cache like `Print`/`Assert`, so the warning re-fires even when no warning sink is installed (`collect_errors=False`). Verified a 3× re-check of one module in a single process warns every time.

Issue #1032, opened 2026-07-16
LALR parser accepts reserved keywords as identifiers in name/binding positions (RD rejects)


jsiek:
## One-sentence summary

The LALR parser accepts reserved keywords (`if`, `define`, `theorem`,
`and`, …) as ordinary identifiers in name / binding positions, whereas the
recursive-descent (RD) parser correctly rejects them — a parser-equivalence
divergence where LALR is too permissive.

## Minimal repro

```
$ echo 'define define = true' > /tmp/x.pf
$ python3 deduce.py --recursive-descent /tmp/x.pf
/tmp/x.pf:1.8-1.13: expected an identifier, not
	"define"
$ python3 deduce.py --lalr /tmp/x.pf
/tmp/x.pf is valid
```

## Observed vs. expected

- **Observed:** under `--lalr`, a program may name a definition, theorem,
  function, union, function parameter, or bound variable after a reserved
  keyword, and it type-checks as valid. Under `--recursive-descent` (the
  default) the same program is a parse error.
- **Expected:** both parsers agree. RD's behavior is the correct one here —
  reserved keywords should not be usable as identifiers.

## Scope

94 keyword strings from `Deduce.lark` are accepted by LALR as a `define`
name but rejected by RD:

> all, and, apply, arbitrary, array, assert, associative, assume, auto,
> bool, by, case, cases, choose, conclude, conjunct, contradict, define,
> else, end, equations, evaluate, expand, export, extensionality, false,
> fn, for, from, fun, generic, ghost, have, help, hiding, if, iff, import,
> in, induction, inductive, injective, into, inverse, inversion, lemma,
> measure, module, not, object, obtain, of, opaque, or, out, postulate,
> predicate, print, private, proof, public, recall, recfun, recursive,
> reflexive, relation, replace, roundtrip, rule, show, simplify, some,
> sorry, source, stop, suffices, suppose, switch, symmetric, target,
> terminates, then, theorem, to, trace, transitive, true, type, union,
> using, var, view, where, with

The divergence is not limited to `define` names. Every binding position is
affected:

| position | example (LALR accepts, RD rejects) |
| --- | --- |
| define name | `define if = true` |
| theorem name | `theorem if: true proof . end` |
| `fun` name | `fun and(x:bool) { x }` |
| `union` name | `union theorem { foo }` |
| `fun` parameter | `fun f(then:bool) { then }` |
| `all`-bound var | `define d = all if:bool. true` |

## Likely root cause

[...]

Issue #854, opened 2026-06-04
Implement imperative verification layer for controlled state and correctness proofs


jsiek:
## Goal

Implement the imperative verification plan in `docs/imperative-verification-plan.md`: add a controlled imperative layer to Deduce for mutable arrays, later object references, dynamic frames, proof slots, and eventually separation-logic resources, while keeping ordinary Deduce terms pure.

Core principle:

> Pure Deduce stays pure. Imperative code lives in a separate computation layer with explicit specifications and explicit proof slots.

## Settled design decisions

- Use `proc` for imperative procedures.
- Use `[T]!` for mutable arrays, distinct from pure arrays `[T]`.
- Start with mutable arrays and local variables; postpone object references to the later object-reference verifier slice.
- Proof slots accept arbitrary proof expressions inline; `proof ... end` is only out-of-line sugar for longer proof terms.
- Call labels expose labeled postconditions as record-like proof facts, e.g. `call push(s, x) as h` followed by `h.valid_post`.
- Object fields do not have their own visibility modifiers; object-level visibility controls field exposure.
- Do not add new automation initially; use existing Deduce mechanisms such as `auto`, `simplify`, and `check_implies`.
- Checked procedures must terminate and prove absence of runtime errors, including out-of-bounds array access.
- Separation assertions are ordinary formulas with resource connectives such as `emp`, `**`, and `|->`; no separate `requires sep` / `ensures sep` clause form.

## Implementation phases

- [x] Phase 1: Add parser/AST support for `proc`, imperative statements, specs, proof slots, observers, objects, resources, and import/export visibility plumbing. No verifier yet. (Parser/AST and declaration plumbing landed: #960–#969, #473. Finishing work tracked by #1107 (allocation parsing), #1108 (unverified-declaration warning), #1109 (CI regressions), #1106 (docs sync). Declarations are accepted but not verified; the verifier arrives in Phase 2.)
- [ ] Phase 2: Verify mutable arrays and local variables with `requires`, `ensures`, `modifies`, `old`, loop `invariant`, `decreases`, and explicit proof slots.
- [ ] Phase 3: Add object references, object fields, allocation, ghost footprints, read-frame checking, and opaque object abstraction.
- [ ] Phase 4: Add separation-logic resources as ordinary formulas with resource-aware proof checking.
- [ ] Phase 5: Build library abstractions and bridge lemmas for stacks, queues, linked lists, circular doubly linked lists, trees, heaps, hash tables, etc.
[...]

PR #1138, opened 2026-07-27
Parse `new Obj<T..>(args)` allocation in imperative RHS positions (closes #1107)

### LINKED ISSUE #1107: Phase 1l: Parse new allocation expressions only in imperative RHS positions
Parent: #854

Phase: 1 finishing work; completes the missing acceptance item from #966.

## Goal

Represent `new Node<T>(args)` as inert imperative AST syntax while continuing to reject `new` in ordinary pure Deduce terms.

## Scope

- Add a dedicated allocation RHS AST node with object name, type arguments, constructor arguments, and source location.
- Extend `imp_rhs` in `Deduce.lark`, `parser.py`, and `rec_desc_parser.py` to accept allocation only after `:=` inside a `proc` body.
- Pretty-print the new node stably.
- Uniquify type arguments and term arguments, but do not perform object allocation or field checking.
- Preserve the existing user-facing rejection for `define x = new Node()` and other pure-term uses.

## Acceptance

- [ ] `var n := new Node<T>(x, next)` parses under both parsers with `--experimental-imperative`.
- [ ] Assignment RHS allocation also parses if the grammar permits it.
- [ ] RD and LALR canonical ASTs match.
- [ ] Pretty-print then reparse preserves the allocation AST.
- [ ] Pure-term `new` remains rejected with the existing focused diagnostic.
- [ ] Add positive round-trip and negative pure-term fixtures.

## Dependencies

None. This is parser/AST-only and can run in parallel with all verifier foundations.

## Out of scope

No allocation semantics, `Ref<T>` checking, field visibility, constructor arity checking, or runtime support. Those remain Phase 3.

jsiek:
## Summary

Phase 1l of the imperative layer (#854): parse `new Obj<T..>(args)` allocation
expressions as inert imperative AST, accepted **only** as a `var`/assignment
right-hand side inside a `proc` body, while continuing to reject `new` in
ordinary pure Deduce terms.

Closes #1107.

## What changed

- **AST** (`abstract_syntax/declarations.py`): new `ImpAlloc` node holding the
  object name, type arguments (`List[Type]`), constructor arguments
  (`List[Term]`), and source location. `ImpVar`/`ImpAssign` RHS unions widened
  to `Term | ImpCallExpr | ImpAlloc`.
- **Grammar** (`Deduce.lark`): the shared `rejected_new_object` production now
  takes real type arguments via a new `new_type_args` rule (`< type_list >`)
  instead of the identifier-only `type_params_opt`, so a compound argument like
  `new Node<List<T>>(...)` parses.
- **LALR** (`parser.py`): a `_imp_rhs_to_ast` helper intercepts the shared
  `rejected_new_object` parse tree in the five imperative RHS positions and
  builds an `ImpAlloc`. Reusing the existing atomic-term shape means no new
  grammar production and therefore no LALR reduce/reduce conflict; pure-term
  `new` still hits the transformer's focused-diagnostic path.
- **RD** (`rec_desc_parser.py`): `parse_imp_rhs` dispatches on the `NEW` token to
  a new `parse_imp_alloc`. Used as a pure term, `new` is still rejected in
  `parse_atom` with the existing message.
- **Uniquify**: `_uniquify_imp_rhs` resolves the type and term arguments; the
  object name is left inert (object resolution / field checking / arity / `Ref<T>`
  are Phase 3).
- **Pretty-print**: `ImpAlloc.__str__` emits `new Obj<T..>(args)` and round-trips
  stably under both parsers.

## Tests

- New positive fixture `test/should-validate/imperative_allocation.pf`, enrolled
  in `EXPERIMENTAL_IMPERATIVE_FILES` and `PARSER_ROUND_TRIP_FILES` (so it is
  checked, RD/LALR equivalence-compared, and pretty-print round-tripped).
- New negative fixture `test/should-error/imperative_new_pure_term.pf` (+ `.err`),
  in the experimental set and in `SHOULD_ERROR_PARSER_EQUIV_SKIP` since both
  parsers reject it.
- `IMPERATIVE_SYNTAX_SOURCE` in `test-deduce.py` gained allocation lines, so the
  in-harness experimental parser test covers RD/LALR AST equality, round-trip,
  and validation.
- Verified locally: `ruff`, `mypy`, `keywords.py`, `reference_grammar.py`, the
  full default `test-deduce.py`, `--equiv`, `--cli`, and both-parser CLI smoke
  checks all pass. Confirmed uniquify resolves the type arg (`T` → `T.sN_M`) and
  term args while leaving the object name unresolved.

## Acceptance checklist (#1107)

- [x] `var n := new Node<T>(x, next)` parses under both parsers with `--experimental-imperative`
- [x] Assignment RHS allocation parses (`a := new Cell(g)`)
- [x] RD and LALR canonical ASTs match
- [x] Pretty-print then reparse preserves the allocation AST
- [x] Pure-term `new` remains rejected with the existing focused diagnostic
- [x] Positive round-trip and negative pure-term fixtures added

## Notes for a resuming human

Nothing outstanding. One pre-existing (not introduced here) RD/LALR divergence
worth knowing: with `--experimental-imperative` **off**, RD demotes `new` to a
plain identifier (per the keyword-contextuality behavior of #1032), so pure-term
`new` yields a generic "expected a statement" parse error under RD versus the
focused message under LALR. With the flag on, both parsers give the focused
message. This is out of scope for #1107 (which targets the flag-on path) and
belongs to the keyword-contextuality umbrella (#1032/#1135).

Resume this session on ginger: `~/deduce-runner/resume.sh 8d4325bf-56d0-4a17-ac06-76300a8b2267 routine/20260727-170202`

### DIFF
--- a/Deduce.lark
+++ b/Deduce.lark
@@ -147,7 +147,7 @@ identifier: IDENT              -> identifier
     | ".0."                                        -> emptyset_literal
     | "emp"                                        -> emp_resource
     | "not" atomic_term                                -> logical_not
-    | "new" IDENT type_params_opt "(" term_list ")"    -> rejected_new_object
+    | "new" IDENT new_type_args "(" term_list ")"      -> rejected_new_object
     | "new"                                            -> rejected_new_term
     | identifier                                   -> term_var
     | "@" atomic_term "<" type_list ">"                -> term_inst
@@ -360,6 +360,12 @@ identifier: IDENT              -> identifier
 
 ?type_params_opt:                                -> empty
    | "<" identifier_list ">"
+
+// Optional type arguments of a `new Obj<T..>(args)` allocation (imperative
+// layer, #854 Phase 1l). Unlike `type_params_opt` these are types, not binding
+// identifiers, so a compound argument such as `new Node<List<T>>(...)` parses.
+?new_type_args:                                  -> empty
+   | "<" type_list ">"
 
 ?constructor: IDENT                                   -> constructor_id
     | IDENT "(" type_list ")"                    -> constructor_apply
--- a/abstract_syntax/declarations.py
+++ b/abstract_syntax/declarations.py
@@ -222,10 +222,27 @@ class ImpCallExpr(AST):
     return result
 
 @dataclass
+class ImpAlloc(AST):
+  # A `new Obj<T..>(args)` allocation used as the right-hand side of a
+  # `var`/assignment inside a `proc` body (Phase 1l, issue #1107). Parsed as
+  # inert syntax: no object allocation, `Ref<T>` checking, field visibility,
+  # or constructor arity checking (those are Phase 3). The pure-term `new`
+  # stays rejected -- allocation is only reachable directly after `:=`.
+  object_name: str
+  type_args: List[Type]
+  args: List[Term]
+
+  def __str__(self) -> str:
+    result = 'new ' + base_name(self.object_name)
+    if self.type_args:
+      result += '<' + ', '.join(str(t) for t in self.type_args) + '>'
+    return result + '(' + ', '.join(str(a) for a in self.args) + ')'
+
+@dataclass
 class ImpVar(ImpStmt):
   name: str
   type_annot: Optional[Type]
-  rhs: Term | ImpCallExpr
+  rhs: Term | ImpCallExpr | ImpAlloc
   ghost: bool = False
 
   def __str__(self) -> str:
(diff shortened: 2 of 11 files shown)


jsiek:
Addressed the P2 review finding: `parse_imp_alloc` in the recursive-descent parser now consumes a bare `IDENT` for the object name (via `consume_token('IDENT', ...)`) instead of the more permissive `parse_identifier()`. This matches the LALR grammar's `"new" IDENT new_type_args ...` production, so `new operator +(x)` and `new __()` are now rejected under both parsers rather than parsing under RD and printing to an unreparseable `new +(x)`. Added `test/should-error/imperative_new_operator_name.pf` as a regression guard.

Also rebased onto current `main`: PR #1136 moved the proc/observer fixtures to `test/should-warn/` (they now emit the Phase 1m unverified-imperative warning), so the positive allocation fixture moved to `test/should-warn/imperative_allocation.pf` with a `.warn` sibling. Full `test-deduce.py`, `--equiv`, ruff, and mypy are green.

PR #1139, opened 2026-07-27
Add a focused imperative-tooling regression lane (closes #1109)

### LINKED ISSUE #1109: Phase 1n: Run imperative parser and LSP symbol regressions in CI
Parent: #854

Phase: 1 finishing work.

## Goal

Ensure the Phase 1 tooling features are continuously tested instead of relying on local-only LSP runs.

## Scope

- Add a narrow CI leg for the imperative declaration symbol-outline test in `test/lsp/test_symbols.py` without running the entire memory-heavy LSP suite.
- Keep the existing experimental-imperative CLI smoke test in an explicit CI job or document exactly which matrix shard owns it.
- Add a regression that verifies every imperative fixture that needs the flag is enrolled in `EXPERIMENTAL_IMPERATIVE_FILES`.
- Add focused negative tests for malformed procedure proof-block entries and duplicate object fields if they are not already covered.

## Acceptance

- [ ] CI fails if `proc`, `observer`, `object`, or `resource` disappears from the LSP outline.
- [ ] CI fails if the feature flag smoke test stops running.
- [ ] CI fails if a tracked imperative fixture is accidentally run without the flag.
- [ ] The new job stays focused and completes without requiring the full LSP suite.

## Dependencies

None. Can run in parallel with all other Phase 1 and Phase 2 work.

## Out of scope

No new LSP features and no broad CI reorganization.

jsiek:
## Summary

Phase 1 of the imperative layer (#854) shipped `proc`/`observer`/`object`/`resource`
declarations, the experimental-imperative parser, and their LSP symbol outline,
but that surface was only exercised by local-only pytest runs or buried inside
the `--cli`/passable shard. This adds a **focused, dependency-light regression
bundle** so CI catches regressions in the Phase 1 tooling directly, addressing
#1109.

`run_imperative_regression` gathers three in-process checks (no memory-heavy LSP
suite, no stdlib bootstrap):

- **LSP symbol outline** still lists `proc`/`observer`/`object`/`resource`.
  Mirrors `test/lsp/test_symbols.py::test_list_symbols_includes_imperative_declarations`
  and uses `lsp.query.list_symbols`, which depends only on `lark` (not on the
  `pygls`/`mcp` server SDKs), so it stays in the focused lane.
- **Experimental-imperative parser smoke test** (`run_experimental_imperative_parser_test`).
- **Flag-enrollment pin**: every `EXPERIMENTAL_IMPERATIVE_FILES` entry genuinely
  needs `--experimental-imperative` — it fails to check without the flag (so a
  stale enrollment turns CI red), and positive fixtures still validate with it.

It runs standalone via `python test-deduce.py --imperative` for focused local
iteration, and also rides on `--cli`, so the existing **passable-1** CI shard
(`--passable --cli`) covers it — no new workflow leg required.

Also adds a negative fixture for a malformed procedure proof-block entry:
`test/should-error/imperative_proc_undefined_label.pf`, whose body's
`assert ... by missing_label` references a label the trailing `proof ... end`
block never defines (rejected with `undefined proof variable missing_label`).

## Acceptance mapping (#1109)

- ✅ CI fails if `proc`/`observer`/`object`/`resource` disappears from the LSP
  outline — `_check_imperative_symbol_outline`.
- ✅ CI fails if the feature-flag smoke test stops running, and the owning shard
  is documented (passable-1) — `run_imperative_regression` under `--cli`.
- ✅ CI fails if a tracked imperative fixture no longer needs the flag —
  `_check_imperative_flag_enrollment`. (The inverse — a *new* imperative
  should-validate file forgotten from the enrollment set — is already caught by
  the `--passable` leg, which would run it without the flag and fail to parse.)
- ✅ Stays focused: in-process, `lark`-only, no full LSP suite, ~2.6s.

## Testing

```
python3 test-deduce.py --imperative   # passes (~2.6s)
python3 test-deduce.py --cli          # passes; now includes the imperative lane
python3 test-deduce.py --errors       # passes (new negative fixture)
python3 test-deduce.py --equiv        # passes
python3 test-deduce.py --passable --cli  # passes
```

`ruff`/`mypy` were unavailable in the container; the new code touches only
`test-deduce.py` (excluded from the mypy config) and adds no line-length or
Pyflakes violations under the repo's `pyproject.toml` ruff selection.

## Note on the dedicated CI matrix leg

The issue's scope also mentions a *dedicated* CI matrix leg. The routine token
lacks `workflow` scope and cannot push `.github/workflows/*` changes, so instead
the lane rides on the existing passable-1 shard (documented in-code). A
maintainer who wants a standalone leg can add this one entry to the
`regression` matrix in `.github/workflows/test_deduce.yml`:

```yaml
          - name: imperative
            args: --imperative
            requirements: core
```

Resume this session on ginger: `~/deduce-runner/resume.sh 9415159b-60e8-4a7b-bd12-f1bf17ddd2e1 routine/20260727-180202`

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -67,6 +67,7 @@ python test-deduce.py --warns          # only test/should-warn (valid + diff vs 
 python test-deduce.py --equiv          # compare RD/LALR ASTs for a curated grammar corpus
 python test-deduce.py --equiv-full     # opt-in: pretty-print round-trip over EVERY accepted-syntax .pf (not in CI)
 python test-deduce.py --parser         # only test/parse (parser-error fixtures)
+python test-deduce.py --imperative     # Phase 1 imperative tooling: LSP outline + parser smoke + flag-enrollment
 python test-deduce.py --site           # generates and checks doc code from gh_pages/doc
 
 # Regenerate the expected stderr fixture for a should-error test
--- a/test-deduce.py
+++ b/test-deduce.py
@@ -1644,41 +1644,58 @@ def _check_imperative_symbol_outline() -> list[tuple[str, str, str]]:
 
 def _check_imperative_flag_enrollment() -> list[tuple[str, str, str]]:
     """Pin that every ``EXPERIMENTAL_IMPERATIVE_FILES`` entry genuinely needs
-    ``--experimental-imperative`` (issue #1109, acceptance box 3). If the flag
-    stopped being load-bearing for a tracked fixture, the enrollment is stale
-    and should be removed; if a positive fixture stopped validating with the
-    flag, the harness would silently run it wrong. All enrolled files are
-    prelude-free, so ``prelude=()`` keeps this off the expensive bootstrap."""
+    ``--experimental-imperative`` (issue #1109, acceptance box 3).
+
+    A fixture is load-bearing only if the flag actually changes its outcome, so
+    we compare the ``(ok, error_message)`` pair for a checked-with-flag run
+    against a checked-without-flag run and require that they *differ*. Merely
+    asserting "fails without the flag" is not enough: a ``should-error`` fixture
+    such as ``imperative_proc_undefined_label.pf`` fails for its own reason with
+    or without the flag, so a no-flag failure alone is no evidence the flag is
+    required. When the two runs are identical the flag is inert and the
+    enrollment is stale. Positive fixtures (should-validate / should-warn /
+    test-imports) must additionally still succeed with the flag. All enrolled
+    files are prelude-free, so ``prelude=()`` avoids the expensive bootstrap."""
     label = "flag-enrollment"
     failures: list[tuple[str, str, str]] = []
-    for path in sorted(EXPERIMENTAL_IMPERATIVE_FILES):
-        if not os.path.exists(path):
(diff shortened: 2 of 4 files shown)


jsiek:
Revised in response to the Codex review (and rebased onto current `main`, which merged #1136/#1138 and expanded the imperative fixture set).

**Enrollment pin now checks that the flag is load-bearing, not merely that the file fails without it.** The reviewer correctly noted that a `should-error` fixture like `imperative_proc_undefined_label.pf` fails for its own reason (`undefined proof variable`) with *or* without the flag, so `not without_flag.ok` proved nothing for negative fixtures. `_check_imperative_flag_enrollment` now compares the `(ok, error_message)` pair of a with-flag run against a without-flag run and flags the entry as stale iff they are **identical** (flag inert). Verified:

- every entry in the (now larger) `EXPERIMENTAL_IMPERATIVE_FILES` set differs on/off → load-bearing;
- a plain non-imperative fixture is correctly detected as inert, so the check has teeth.

Also quiet+redirect the newly-enrolled `should-warn` fixtures' unverified-declaration warnings so the lane's output stays clean.

`--imperative`, `--cli`, `--errors`, and `--warns` all pass locally after the rebase.

Issue #1144, opened 2026-07-27
Spurious 'already defined' warning on legitimate nested imperative var shadowing


jsiek:
**Is this intentional? If so, close.**

Legitimate lexical shadowing of an imperative local `var` in a nested block emits a spurious `WARNING: ... is already defined`, even though nested-block shadowing is a supported feature of the imperative layer (issue #1113 explicitly requires "Nested shadowing behaves consistently" and that "nested block locals must not leak").

## Minimal repro

```
// shadow.pf
proc demo(flag: bool)
{
  var a := flag
  if flag {
    var a := flag
  }
}
```

```sh
python3 deduce.py --experimental-imperative --no-stdlib shadow.pf
```

## Observed

```
shadow.pf:5.5-5.18: WARNING: a is already defined
... proc 'demo' is accepted but not verified ...
shadow.pf is valid
```

## Expected

No `is already defined` warning: the inner `var a` is a distinct block-local that shadows the outer `a` in a nested scope, which uniquify already alpha-renames to a distinct name. (A re-declaration of `a` in the *same* block is a different case and could reasonably warn.)

## Source

The warning comes from `abstract_syntax/declarations.py::overwrite` (called by `_uniquify_imp_stmt` for `ImpVar`): a nested block receives a `copy_dict(env)` that still contains the outer binding, so `overwrite` sees the name already present and warns. Surfaced while adding Phase 2d body type-checking (issue #1113); the type-checker itself scopes shadowing correctly, but the uniquify-side warning fires first.

Refs #854.

jsiek:
Taking this — labeling in progress. Fix: thread a per-block `declared` set through the imperative-body uniquify walk so `overwrite` only warns on a *same-block* re-declaration; shadowing an outer-block `var` (a distinct alpha-renamed local) no longer warns.

Issue #1122, opened 2026-07-27
Phase 2m: Verify loop decreases clauses and local_counter


jsiek:
Parent: #854

Phase: 2 total-correctness verification.

## Goal

Require every checked loop to terminate by proving a well-founded measure is nonnegative and strictly decreases on each iteration.

## Scope

- Type-check the loop `decreases` term against the chosen natural/unsigned measure discipline.
- Capture the measure before the body and generate nonnegativity and strict-decrease obligations under invariant plus loop condition.
- Reject missing decreases clauses unless a documented trivial-loop exception exists.
- Attach failures to the `decreases` annotation.
- Reuse ordinary Deduce arithmetic facts and proof automation; add no new solver.

## Acceptance

- [ ] Add `local_counter.pf` as a validating fixture.
- [ ] Add `bad_decreases.pf` as an error fixture showing that an increasing measure fails.
- [ ] A constant measure fails.
- [ ] A measure with the wrong type fails during type checking.
- [ ] Loop invariant and decreases obligations remain distinct in diagnostics.

## Dependencies

Depends on #1121.

## Parallelization

Can run in parallel with proof-slot support and array-specific work after the loop rule is stable.

## Out of scope

No recursive procedure termination and no lexicographic or user-defined well-founded orders.

Issue #1117, opened 2026-07-27
Phase 2h: Type mutable-array length and reads and emit bounds obligations


jsiek:
Parent: #854

Phase: 2 mutable-array read support.

## Goal

Allow specifications and procedure expressions to inspect [T]! arrays safely without changing the semantics of pure arrays [T].

## Scope

- Type length(a) for a mutable-array handle and define length as immutable for the lifetime of the handle in Phase 2.
- Type a[i] as the mutable-array element type.
- Require an unsigned/integer index type consistent with existing Deduce arrays.
- Emit a source-located obligation i < length(a) for every imperative mutable-array read.
- Avoid emitting duplicate bounds goals for the same source access within one verification pass.
- Keep pure-array ArrayType behavior unchanged.

## Acceptance

- [ ] In-bounds reads verify when a precondition supplies the bound.
- [ ] Missing bounds evidence produces the target array-access diagnostic.
- [ ] Wrong index and element expectations are rejected by type checking.
- [ ] Pure-array regression fixtures are unchanged.
- [ ] Unit tests distinguish ArrayType and MutableArrayType paths.

## Dependencies

Depends on #1111 and #1112.

## Parallelization

Can run in parallel with #1113-#1116. It unlocks mutable-array writes and read-only algorithms. Ghost-flow checks for uses of array reads remain owned by #1114.

## Out of scope

No writes, allocation, aliasing, frame checking, or observer preservation.

jsiek:
Claude is taking this issue in progress (routine run): Phase 2h — typing mutable-array `length`/reads and emitting source-located bounds obligations, in parallel with the #1113–#1116 chain per the issue's parallelization note.

Issue #1116, opened 2026-07-27
Phase 2g: Verify if, assert, and assume control flow


jsiek:
Parent: #854

Phase: 2 control-flow verification.

## Goal

Extend straight-line verification to finite branching and proof-only assumptions.

## Scope

- Type-check imperative `if` conditions as `bool`.
- Verify both branches against the same continuation and add the condition or its negation to branch givens.
- Support `if` without `else` by treating the missing branch as `skip`.
- Generate an obligation for `assert P`, then add `P` to subsequent givens.
- Add `assume P` to subsequent givens only in ghost/proof state and ensure it has no runtime effect.
- Diagnose missing returns when only some branches return.

## Acceptance

- [ ] Positive and negative branch examples verify/fail at the expected locations.
- [ ] Facts asserted or assumed are available only after the statement.
- [ ] Both branches are checked, including unreachable-looking branches unless proved unreachable.
- [ ] Partial-return and wrong-condition-type fixtures are covered.
- [ ] Generated obligations have deterministic given ordering.

## Dependencies

Depends on #1114 and #1115.

## Parallelization

Can run in parallel with mutable-array read typing (#1117) and proof-slot resolution once their foundations are ready.

## Out of scope

No loops, heap writes, calls, or path-sensitive alias analysis.

jsiek:
Working on this now (routine). Marking in progress — deps #1114 and #1115 have merged, so straight-line control-flow verification (`if`/`assert`/`assume`) can extend `_proc_verifiable`/`proc_obligations`.

PR #1145, opened 2026-07-27
Type-check imperative proc bodies: locals, assignments, returns (closes #1113)

### LINKED ISSUE #1113: Phase 2d: Type-check local variables, assignments, and returns
Parent: #854

Phase: 2 statement typing.

## Goal

Type-check straight-line imperative blocks without yet proving specifications.

## Scope

- Type-check annotated and inferred `var` declarations.
- Track block-local mutable bindings with lexical scope; nested block locals must not leak.
- Type-check local-variable assignment and reject assignment to unknown or non-local names.
- Type-check `return e` against the declared result type and reject value returns from procedures with no result.
- Diagnose missing returns conservatively for straight-line bodies; later control-flow issues extend path analysis.
- Preserve source locations and uniquified local names.

## Acceptance

- [ ] Annotated and inferred locals accept matching initializers.
- [ ] Wrong initializer and assignment types produce focused diagnostics.
- [ ] Nested shadowing behaves consistently under both parsers.
- [ ] Return type mismatches and invalid returns are rejected.
- [ ] Add positive and negative fixtures to `test/imperative`.

## Dependencies

Depends on #1111.

## Parallelization

Can proceed while #1112 is implemented. Mutable-array expression typing can start from #1111 in parallel.

## Out of scope

No weakest preconditions, branching path joins, ghost noninterference, array writes, calls, loops, or proof checking.

jsiek:
## Summary

Phase 2d of the imperative verification layer (#854): type-check straight-line
procedure bodies without yet proving specifications.

`type_check_proc_body` runs in `type_check_stmt` for `ProcDecl` and checks:

- **annotated and inferred local `var` declarations** — an annotation is
  checked against the initializer; an unannotated `var` synthesizes its type
  from the initializer;
- **assignment to local variables** — the target must be a block-local
  `var`/parameter; assigning to an unknown name or a global is rejected with a
  focused diagnostic;
- **`return e`** — checked against the declared result type; a value `return`
  in a procedure with no result type is rejected;
- **nested `if`-block lexical scope** — block-locals do not leak into the
  enclosing block (`Env` is functional, so nested blocks type-check against a
  copy);
- **conservative missing-return** — a straight-line body that declares a
  result type but can fall off the end is rejected; the check is sound (only
  fires when it is certain the body can complete without returning).

### Out of scope (deferred to later slices)

Constructs not yet modeled by the verifier — procedure calls, allocations,
mutable-array reads/writes, field writes, `assert`/`assume`, and loops
(#1116–#1122) — are deferred **wholesale**: a body using any of them is left
untouched and keeps emitting the Phase 1m "accepted but not verified" warning
(#1108) rather than risk a spurious type error. A mutable-array (`[T]!`)
parameter defers the whole body, since `a[i]` reads are not typeable until
#1117. No specifications are proved here.

## Tests

New fixtures under `test/imperative`, run under both parsers by the
`--imperative` lane (which also rides `--passable --cli` in CI):

- `should-validate/proc_body_locals.pf` — positive path (annotated + inferred
  locals, assignment, nested-`if` scope, matching `return`).
- `should-error/` negatives: wrong initializer type, return-type mismatch,
  value return in a void proc, assignment to an unknown name, assignment to a
  global (non-local), missing return, and a nested-block local that does not
  leak.

`return_in_void` has a `.pf.err.lalr` sibling for the known RD/LALR return
location-span divergence.

Ran locally: `test-deduce.py --imperative --warns --errors --passable --equiv
--lib`, `ruff check .`, `mypy .`, and the token/grammar checks — all green.

## Bug filed in passing

- #1144 — spurious `already defined` warning on legitimate nested `var`
  shadowing (pre-existing uniquify behavior surfaced while scoping this slice).

Closes #1113. Refs #854.

Resume this session on ginger: `~/deduce-runner/resume.sh 62f47fd3-d10e-4e38-8096-6f4732a69f79 routine/20260727-230202`

### DIFF
--- a/checker_pipeline.py
+++ b/checker_pipeline.py
@@ -22,11 +22,13 @@ from abstract_syntax import (
     All, And, Array, ArrayGet, Assert, Associative, Auto, Bool,
     Call, Conditional, Constructor, Declaration, Define, Env, Export,
     Formula, FunCase, FunctionType, GenRecFun, Generic, GenericUnknownInst,
-    Hole, IfThen, Import, Inductive, Lambda, MakeArray, Module, ObjectDecl,
+    Hole, IfThen, ImpAlloc, ImpAssign, ImpCallExpr, ImpIf, ImpReturn, ImpStmt,
+    ImpVar, Import, Inductive, Lambda, LValueVar, MakeArray, Module,
+    MutableArrayType, ObjectDecl,
     ObjectField, ObserverDecl, Omitted, Or, OverloadType, OverloadedVar, PSorry, PVar,
     PatternBool, PatternCons, Postulate, Predicate, Print, ProcDecl, ProcParam,
     ProcSpec, RecFun, ResolvedVar, ResourceDecl, Rule, Some,
-    Statement, Switch, SwitchCase, TAnnote, TLet, Term, TermInst, Theorem,
+    Statement, Switch, SwitchCase, TAnnote, TermBinding, TLet, Term, TermInst, Theorem,
     Trace, Type, TypeAlias, TypeInst, TypeType, Union, Var, VarRef, VerboseLevel,
     ViewDecl, ViewRecFun, alpha_equiv, base_name, callable_name,
     check_post_typecheck_invariants, find_file, full_reduce, mkEqual,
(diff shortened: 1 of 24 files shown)


jsiek:
Thanks — both Codex findings were real bugs; fixed in 2799655.

**P1 (defer call/new RHSs in local vars):** `_imp_stmt_unmodeled` matched `ImpVar` and `ImpAssign` with a shared positional pattern, but `ImpVar`'s fields are `(location, name, type_annot, rhs, ghost)` while `ImpAssign`'s are `(location, lhs, rhs)`, so the shared pattern bound `rhs` to `ImpVar.type_annot`. A `var x := new C(..)` / `var x := call f(..)` slipped past the deferral gate and reached the term checker as a hard `cannot synthesize a type for new C(x)` error. Switched to field access (`s.rhs`/`s.lhs`). Confirmed repro `proc f() -> bool { var a := new Cell(x)  return x }` now defers with the Phase 1m warning; regression fixture `should-validate/proc_body_defer_alloc.pf`.

**P2 (recognize earlier guaranteed returns):** `_block_always_returns` only inspected `stmts[-1]`, so an early `return` (or an `if`/`else` whose branches both return) followed by dead code was wrongly flagged as a possible fall-through. Now any statement that always returns makes the whole block always return (everything after is unreachable). Regression fixture `should-validate/proc_body_if_returns.pf` (terminal `if`/`else` return, no trailing `return`).

Also added `should-error/proc_body_assign_type.pf` after an earlier question about assignment RHS type-checking. `--imperative`, `--warns`, `--errors`, `ruff`, and `mypy` all green.

Issue #1115, opened 2026-07-27
Phase 2f: Verify straight-line local procedures with requires and ensures


jsiek:
Parent: #854

Phase: 2 core verification.

## Goal

Verify procedures containing only local declarations, local assignments, assertions, and a final return by computing weakest preconditions over local state.

## Scope

- Treat `requires` clauses as entry givens and conjoin repeated clauses.
- Treat repeated `ensures` clauses as exit goals with `result` substituted.
- Implement symbolic substitution for local declarations and assignments.
- Generate assertion obligations and continue with the asserted fact after discharge.
- Reject `assume` in this slice or delegate it to #1116; do not silently treat it as proven.
- Handle empty bodies and procedures without a return value.
- Remove the unchecked warning from #1108 only for procedures entirely supported by this slice.

## Acceptance

- [ ] A straight-line identity/counter-without-loop procedure verifies.
- [ ] A wrong return value fails at the postcondition.
- [ ] A false assertion reports its own source location.
- [ ] Multiple `requires` and `ensures` clauses retain their source locations.
- [ ] Unit tests pin the generated weakest-precondition formulas.

## Dependencies

Depends on #1112 and #1113.

## Parallelization

Can run in parallel with #1114 and mutable-array expression typing. It unlocks conditionals, arrays, calls, and loops.

## Out of scope

No branches, loops, mutable heap, procedure calls, proof slots, `old`, or runtime execution.

jsiek:
Starting work on this (Phase 2f: straight-line proc verification with requires/ensures). Marking in progress.

Issue #1112, opened 2026-07-27
Phase 2c: Add a source-located imperative verification-obligation API


jsiek:
Parent: #854

Phase: 2 verifier foundation.

## Goal

Provide one reusable representation and discharge path for all imperative proof obligations before individual statement rules are implemented.

## Scope

- Add an `ImperativeObligation` or equivalent data type containing source location, goal formula, ordered givens, obligation kind, and optional attached proof.
- Define stable kinds for precondition, postcondition, assertion, array bounds, frame, loop entry, loop preservation, loop exit, and decreases obligations.
- Bridge obligation discharge to existing Deduce proof checking and existing automation such as `auto`, `simplify`, and `check_implies`.
- When no proof discharges a goal, report the normal `Goal:` and `Givens:` presentation at the source annotation that created it.
- Keep obligation construction pure/testable so later verifier slices can unit-test formulas without running the entire CLI.

## Acceptance

- [ ] Unit tests construct and discharge a trivial obligation.
- [ ] A false obligation produces a source-located incomplete-proof diagnostic.
- [ ] Existing proof checker behavior is reused rather than copied.
- [ ] Obligation kinds are visible in diagnostics or debug output.
- [ ] No generated global names are exposed to users.

## Dependencies

Depends on #1110.

## Parallelization

Can be implemented in parallel with #1111. It unlocks straight-line verification, bounds checks, loops, and proof slots.

## Out of scope

No statement-specific VC generation, symbolic heap, or LSP code actions.

jsiek:
Starting work on this (Phase 2c: source-located imperative verification-obligation API). Adding a pure/testable `ImperativeObligation` representation with a single discharge path that bridges to the existing proof checker (attached proofs via `check_proof_of`) and automation (`reduce`/auto rules + `check_implies`), reporting the standard source-located `Goal:`/`Givens:` incomplete-proof diagnostic when nothing discharges the goal.

Issue #1111, opened 2026-07-27
Phase 2b: Register and type-check procedure signatures and contracts


jsiek:
Parent: #854

Phase: 2 verifier foundation.

## Goal

Make a `proc` a typed callable declaration with a checked public contract, while leaving statement execution and verification to later slices.

## Scope

- Introduce a procedure-signature representation containing type parameters, ordinary and ghost parameter types, optional return type, and ordered spec clauses.
- Register procedure signatures in `Env` so later calls can resolve them without treating them as pure functions.
- Type-check parameter types, return types, `requires` and `ensures` formulas, frame expressions at the structural level, and `decreases` terms.
- Bind `result` only in postconditions of procedures with a return type.
- Reject duplicate procedure names and malformed or duplicate labelled postconditions with focused diagnostics.
- Keep procedure names unavailable to ordinary pure-term call syntax.

## Acceptance

- [ ] Generic and non-generic procedure signatures type-check.
- [ ] `requires` and `ensures` must have type `bool`.
- [ ] `result` is accepted only where defined and has the declared return type.
- [ ] A pure `fun` or `define` cannot call a procedure as a normal term.
- [ ] Imports expose procedure signatures without bodies.
- [ ] Add positive and negative fixtures to the imperative harness.

## Dependencies

Depends on #1110.

## Parallelization

Can be implemented in parallel with #1112. It unlocks local statement typing and mutable-array expression typing.

## Out of scope

No body checking, weakest preconditions, call statements, frames, proof slots, or runtime behavior.

jsiek:
Working on this now (Phase 2b: register and type-check procedure signatures and contracts). Marking in progress.

PR #1150, opened 2026-07-28
Type mutable-array length/reads and emit bounds obligations (closes #1117)

### LINKED ISSUE #1117: Phase 2h: Type mutable-array length and reads and emit bounds obligations

jsiek:
## Phase 2h: type mutable-array `length`/reads and emit bounds obligations

Closes #1117 (parent umbrella #854). Depends only on the already-merged
#1111 (proc signatures/contracts) and #1112 (obligation API), so this runs in
parallel with the #1113–#1116 chain exactly as the issue's parallelization note
describes.

### What changed

**Type checking (`checker_types.py`)**
- `a[i]` on a mutable `[T]!` handle now type-checks to the element type `T`,
  mirroring the pure `[T]` path, and requires a `UInt` index (`length(a)` is
  `UInt` and `<` compares homogeneous pairs, so only a `UInt` index can form a
  provable `i < length(a)` bound). The pure `ArrayType` path is left
  byte-for-byte unchanged, so existing pure-array behavior (which never checked
  the index type) is preserved.
- `length(a)` on a mutable `[T]!` handle type-checks to the array-length type
  (`UInt`) via a new symbolic `ArrayLength` term node. The result type is read
  off the in-scope `length` function's return type rather than fabricated, so
  `length(a)` on an array agrees with `length` on a list. Any other `length(...)`
  call — including on a pure `[T]` array — falls back to ordinary list-`length`
  resolution. The intercept is applied in both `type_synth_term` and
  `type_check_term` so it works in synthesis, call-argument, and equality
  positions.

**Obligation API (`imperative_verifier.py`)**
- `array_bounds_goal(read)` builds the source-located `i < length(a)`
  bounds goal for a mutable-array read, using the same post-typecheck
  `ResolvedVar` constructor idiom `mkEqual` uses for `=`.
- `ArrayBoundsObligations` collects `ARRAY_BOUNDS` obligations for the reads
  seen in one verification pass and deduplicates repeated reads of the same
  source access, so a given `a[i]` yields at most one bounds goal.

**New AST node**
- `ArrayLength` in `abstract_syntax/terms.py`. It is produced only by the type
  checker (never parsed), so both parsers still see the surface `length(a)` as
  an ordinary `Call` and parser-equivalence/round-trip coverage is unaffected.
  It has no runtime reduction (an array's length is immutable for the handle's
  lifetime in Phase 2). The proof-rewrite walkers (`count_marks` / `find_mark` /
  `replace_mark` / `rewrite_aux`) are taught about it so `replace`/`simplify`/
  rewrite tactics work on formulas mentioning `length(a)`. Scoping the `length`
  intercept to `[T]!` keeps `ArrayLength` confined to procedure specs and bounds
  obligations, which never reach the recursive-call or compiler walkers that
  pure runtime terms use.

### Acceptance criteria

- [x] In-bounds reads verify when a precondition supplies the bound — unit test
  discharges the obligation against a matching `requires i < length(a)` given.
- [x] Missing bounds evidence produces the array-bounds diagnostic — unit test.
- [x] Wrong index and element expectations are rejected by type checking —
  `should-error` fixture (bool index) plus unit tests; element mismatch is
  rejected by the existing `=` check over the read's element type.
- [x] Pure-array regression fixtures are unchanged — full suite green.
- [x] Unit tests distinguish `ArrayType` and `MutableArrayType` paths —
  `test/unit/test_array_bounds.py`.

### Scope boundary

This slice delivers the **typing** and the **obligation-construction/dedup**
machinery. The pass that walks a verified procedure body and feeds each read
into `ArrayBoundsObligations.record(...)` is straight-line VC generation
(#1115+), which does not exist on `main` yet; the API here is the tool that
slice will call. Everything in this PR is exercised end-to-end through the
type checker (specs are type-checked today via #1111) and at the unit level.

### Review response (Codex on `4f39468`)

- **Rewrite walkers**: added `ArrayLength` cases to `count_marks`, `find_mark`,
  `replace_mark`, and `rewrite_aux` so rewriting/simplifying a formula with
  `length(a)` no longer hits `internal_error`.
- **Index type**: mutable-array indexes are now restricted to `UInt` (was
  `UInt`/`Nat`/`Int`), so a read always admits a provable `i < length(a)` bound.
- **Intercept scope**: `length(a)` typing is restricted to mutable `[T]!` only,
  keeping `ArrayLength` out of the recursive-call/compiler walkers.

### Testing

- `python3 test-deduce.py` (full default incl. parser equivalence) — green.
- `python3 test-deduce.py --imperative` — green (both parsers; flag enrollment).
- `ruff check .`, `mypy .`, `keywords.py`, `reference_grammar.py` — all green.
- New `test/unit/test_array_bounds.py` — 9/9 (run manually; pytest not present
  in this container).
- LALR parity smoke-checked on both new fixtures.

Resume this session on ginger: `~/deduce-runner/resume.sh ce1bb127-0e51-4624-8b17-22ff0279807e routine/20260728-040201`

### DIFF
--- a/abstract_syntax/rewrite.py
+++ b/abstract_syntax/rewrite.py
@@ -104,6 +104,8 @@ def count_marks(formula: AST) -> int:
       return sum([count_marks(elt) for elt in elements])
     case MakeArray(_, _, subject):
       return count_marks(subject)
+    case ArrayLength(_, _, subject):
+      return count_marks(subject)
     case _:
       internal_error(formula.location, 'in count_marks function, unhandled ' + str(formula))
 
@@ -164,6 +166,8 @@ def find_mark(formula: AST) -> None:
       for elt in elements:
           find_mark(elt)
     case MakeArray(_, _, subject):
+      find_mark(subject)
+    case ArrayLength(_, _, subject):
       find_mark(subject)
     case _:
       internal_error(formula.location, 'in find_mark function, unhandled ' + str(formula))
@@ -238,6 +242,8 @@ def replace_mark(formula: Term | SwitchCase, replacement: Term) -> Term | Switch
                    [replace_mark(elt, replacement) for elt in elements])
     case MakeArray(loc2, tyof, subject):
       return MakeArray(loc2, tyof, replace_mark(subject, replacement))
+    case ArrayLength(loc2, tyof, subject):
+      return ArrayLength(loc2, tyof, replace_mark(subject, replacement))
     case _:
       internal_error(formula.location, 'in replace_mark function, unhandled ' + str(formula))
 
@@ -436,7 +442,11 @@ def rewrite_aux(loc: Meta, formula: Term | SwitchCase, equation: Formula | AutoR
     case MakeArray(loc2, tyof, subject):
       return MakeArray(loc, tyof,
                        rewrite_aux(loc, subject, equation, env, depth - 1))
-  
+
+    case ArrayLength(loc2, tyof, subject):
+      return ArrayLength(loc, tyof,
+                         rewrite_aux(loc, subject, equation, env, depth - 1))
+
     case TLet(loc2, tyof, var, rhs, body):
       return TLet(loc2, tyof, var, rewrite_aux(loc, rhs, equation, env, depth - 1),
                   rewrite_aux(loc, body, equation, env, depth - 1))
--- a/abstract_syntax/terms.py
+++ b/abstract_syntax/terms.py
@@ -1666,6 +1666,24 @@ class ArrayGet(Term):
     return ArrayGet(self.location, self.typeof, subject_red, position_red)
 
 @dataclass
+class ArrayLength(Term):
+  # `length(a)` for a mutable-array handle (or a pure array). Produced only
+  # by the type checker when `length` is applied to an array-typed argument
+  # (issue #1117, Phase 2h); it is never parsed, so both parsers still see
+  # the surface `length(a)` as an ordinary `Call`. In Phase 2 an array's
+  # length is immutable for the lifetime of the handle, so this node has no
+  # runtime reduction -- the default `_map_children` walk only reduces the
+  # `subject`, leaving `length(a)` symbolic for use inside specifications and
+  # array-bounds obligations.
+  subject: Term
+
+  def __eq__(self, other: object) -> bool:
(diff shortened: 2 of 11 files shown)


jsiek:
Thanks — both P2 comments addressed in `2add2f8`:

1. **Rewrite walkers**: `ArrayLength` is now handled in `count_marks`, `find_mark`, `replace_mark`, and `rewrite_aux`, so `replace`/`simplify`/rewrite on a formula mentioning `length(a)` no longer falls through to `internal_error` (verified: `count_marks`/`find_mark`/`replace_mark` recurse into the subject).
2. **Index type**: mutable-array indexes are restricted to `UInt` (was UInt/Nat/Int), matching the `UInt`-typed `length(a)` and the homogeneous `<` overloads, so a read always admits a provable `i < length(a)` bound.

I also scoped the `length(a)` intercept to mutable `[T]!` handles only (pure `[T]` length still falls back to list resolution as before), which keeps `ArrayLength` out of the recursive-call/compiler walkers entirely. Full suite + ruff/mypy green.

Issue #1114, opened 2026-07-27
Phase 2e: Enforce ghost-variable noninterference in procedure bodies


jsiek:
Parent: #854

Phase: 2 statement typing.

## Goal

Make ghost locals and ghost parameters proof-only so later compilation can erase them without changing runtime behavior.

## Scope

- Track whether each procedure parameter and local variable is ghost or runtime.
- Permit ghost expressions in specifications, invariants, assertions, assumptions, and proof terms.
- Reject a runtime local initializer, runtime assignment, array index/value, branch condition, call runtime argument, or return value that depends on ghost data.
- Permit runtime data to flow into ghost variables.
- Produce a diagnostic that names the ghost binding and the runtime context it would influence.

## Acceptance

- [ ] Ghost-to-runtime flows are rejected for each supported straight-line context.
- [ ] Runtime-to-ghost flows validate.
- [ ] Pure proof/specification uses of ghost values validate.
- [ ] Add table-driven unit tests for the dependency predicate and `.pf` regression fixtures.

## Dependencies

Depends on #1113.

## Parallelization

Can run in parallel with straight-line VC generation (#1115) and mutable-array read typing after local binding metadata is stable.

## Out of scope

No code generation or physical erasure; Phase 6 performs erasure.

jsiek:
Picking this up (Phase 2e: ghost-variable noninterference). Marking in progress.

Issue #1120, opened 2026-07-27
Phase 2k: Implement entry-state old and unchanged formulas for arrays


jsiek:
Parent: #854

Phase: 2 specification state snapshots.

## Goal

Let postconditions compare current mutable-array state with procedure-entry state.

## Scope

- Add dedicated AST/elaboration support for `old(e)` in imperative specifications instead of treating `old` as an ordinary unresolved function.
- Restrict `old` to specification/proof contexts where an entry snapshot exists.
- Evaluate mutable-array reads inside `old` against the entry heap version.
- Implement `unchanged(F)` as a checked shorthand over the supported Phase 2 array frames.
- Ensure a callee-level pre-call snapshot can be added later without conflating it with procedure entry.
- Update grammar, both parsers, pretty-printing, highlighting, and reference grammar as needed.

## Acceptance

- [ ] Array swap postconditions can mention `old(a[i])` and `old(a[j])`.
- [ ] `old` outside a valid specification context is rejected clearly.
- [ ] `unchanged(a[i])` succeeds/fails according to symbolic writes.
- [ ] RD/LALR equivalence and round-trip coverage include both forms.
- [ ] Pure Deduce identifiers named `old` or `unchanged` retain the documented flag-off behavior.

## Dependencies

Depends on #1117 and #1118.

## Parallelization

Can run in parallel with modifies-frame enforcement (#1119).

## Out of scope

No call-site `old` semantics, object fields, footprints, or separation resources.

PR #1160, opened 2026-08-07
Verify straight-line procedures with requires/ensures (closes #1115)

### LINKED ISSUE #1115: Phase 2f: Verify straight-line local procedures with requires and ensures

jsiek:
## Phase 2f: verify straight-line procedures with `requires`/`ensures`

Closes #1115. Part of the imperative verification layer (issue #854).

### What this does

`verify_proc` (in `check_proofs`) now *verifies* a procedure whose body is
entirely straight-line — local `var` declarations, assignments to a local,
`assert`, and a final `return`, all with ordinary-term right-hand sides — by
forward symbolic execution:

- **`requires` → entry givens.** Every `requires` clause becomes a given;
  repeated clauses are conjoined by the obligation's `givens_formula`.
- **Symbolic state.** A `state` map threads each local variable to its current
  value as a term over the parameters, so an `assert` formula or a returned
  expression substitutes down to a goal that mentions only parameters. A
  reassigned local reflects its *last* write.
- **`assert` → obligation + fact.** Each `assert P` raises an ASSERTION
  obligation at its own location, and then `P` is added to the downstream
  givens (available to later asserts and the postconditions).
- **`ensures` → exit goals.** Each `ensures` clause becomes its own
  POSTCONDITION obligation, anchored at that clause, with the returned value
  substituted for `result`. A void procedure that falls off the end checks its
  postconditions with `result` unbound.

The verification-condition generation lives in a pure, unit-testable
`proc_obligations(decl, env)` that returns the obligations without discharging
them; `verify_proc` discharges what it returns via the existing
`imperative_verifier` obligation API (`reduce` + `check_implies`, or an
attached inline `assert ... by` proof).

Bodies this slice does not model — branches (`if`), loops (`while`), `assume`,
procedure calls, allocations, and mutable-array reads/writes, or any
mutable-array parameter — stay **deferred** and keep the Phase 1m
"accepted but not verified" warning (issue #1108). `assert` statements are now
also type-checked (as `bool`) in the Phase 2d body pass so their formulas are
well-formed before verification.

### Acceptance (issue #1115)

- ✅ A straight-line identity/counter procedure verifies
  (`test/imperative/should-validate/proc_verify.pf`).
- ✅ A wrong return value fails at the postcondition
  (`proc_wrong_return.pf`).
- ✅ A false assertion reports its own source location
  (`proc_false_assert.pf`).
- ✅ Multiple `requires`/`ensures` clauses retain their source locations
  (unit test + `proc_verify.pf::keep_both`).
- ✅ Unit tests pin the generated weakest-precondition formulas
  (`test/unit/test_imperative_verify.py`, 10 tests).

### Fixture churn (procedures that now verify)

- `imperative_false_proc.pf` retired — it *was* the "false postcondition"
  demo, now the canonical `proc_wrong_return.pf` should-error fixture; removed
  from `EXPERIMENTAL_IMPERATIVE_FILES`.
- `proc_declaration.pf`, `proc_declarations.pf`, `proc_signatures.pf` lose the
  Phase 1m warnings for the straight-line procedures the verifier now
  discharges (`.warn` fixtures regenerated; `demo_nudge`'s unprovable
  `b or not b` tautology changed to a provable `b = b`, since the auto engine
  does case-analysis-free implication only).

### Testing

- `python3 test-deduce.py` — all lanes green (lib, should-validate ×2 parsers,
  should-error, should-warn, prelude, parser equivalence, round-trip).
- `python3 test-deduce.py --imperative` — green (both parsers).
- `python3 -m ruff check .` and `python3 -m mypy .` — clean.
- `test/unit/test_imperative_verify.py`, `test_imperative_obligation.py`,
  `test_array_bounds.py` — pass (run directly; pytest is not CI-gated).

### Notes for a human picking this up

- Phase 2e (#1114, PR #1159, ghost noninterference) was still unmerged when
  this was written; it and 2f both extend `type_check_proc_body` /
  `_type_check_imp_stmt`, so a merge may touch adjacent lines there.
- The Phase 1m warning is retired **only** for procedures 2f fully models; 2g
  (#1116, `if`/`assert`/`assume` control flow) is the next slice and will
  extend `_proc_verifiable` to branches.

Resume this session on ginger: `~/deduce-runner/resume.sh c6de4ca9-0320-43b9-b2ae-c7f421e78c8c routine/20260807-170202`


jsiek:
Thanks @chatgpt-codex-connector — both findings are real and are now fixed in `40a3889`. In both cases the right move for this slice is to **defer** the procedure (keep the Phase 1m "not verified" warning) rather than verify it, since each touches semantics that are explicitly out of scope for #1115.

**P1 — assigning to a parameter (soundness).** Confirmed: `requires not x; ensures not x { x := true }` was wrongly reported valid, because the postcondition was checked against the *entry* value of `x` while the body had reassigned it. Parameters are declared as locals so the assignment type-checks, but without an entry-state/`old` snapshot (the later slice #1120) a postcondition mentioning a reassigned parameter is ambiguous between its entry and exit value. `_proc_verifiable` now defers any procedure that assigns to a parameter, so the deliberately-false contract stays on the unverified/warning path instead of being accepted. Regression: `test/imperative/should-validate/proc_defer_param_assign.pf` (+ `.warn`) plus a `_proc_verifiable` unit test.

**P2 — `assert P by <slot>` proof slots.** Confirmed: such a procedure reported `could not find given: slot…` once it became straight-line, because the out-of-line `proof … end` block's slot bindings are never installed. Proof slots are out of scope for this slice, so `_proc_verifiable` now defers any procedure that declares a `proof` block. An inline `assert P by <proof>` with no proof block is still verified. Regression: `test/imperative/should-validate/proc_defer_proof_slot.pf` (+ `.warn`) plus a `_proc_verifiable` unit test.

Both deferrals will be lifted by later slices (#1120 for `old`/exit-state, and the proof-slot resolution slice), which is where the entry-vs-exit and slot-binding semantics belong. Full `test-deduce.py`, `--imperative`, `ruff`, and `mypy` are green after the change.

PR #1161, opened 2026-08-07
Guard that every user-facing Deduce.lark rule is documented (refs #473)

jsiek:
## Guard that every user-facing Deduce.lark rule is documented

Refs #473 (part B: Reference.md ↔ LALR grammar sync).

### Problem

`gh_pages/scripts/reference_grammar.py` verifies that every documented
`deduce-grammar` fragment agrees with `Deduce.lark` — but only in one
direction. Nothing checked the reverse: a new **user-facing** grammar rule
could be added to `Deduce.lark` and never documented in `Reference.md` /
`ImperativeReference.md`, and the checker would stay green. That is exactly
the reference↔grammar drift #473 exists to catch, just in the
`Deduce.lark → reference` direction.

Today 81 of the 91 lowercase rules in `Deduce.lark` are documented in checked
blocks; the other 10 are all internal (start rule, list containers,
passthrough/precedence helpers, lexical wrappers).

### Change

Add a reverse-direction coverage guard in `main()`:

- Every lowercase `Deduce.lark` rule must appear in some `deduce-grammar`
  block **or** in `INTERNAL_UNDOCUMENTED_RULES`, a new explicit allowlist of
  the 10 internal/structural rules, each with a one-line rationale.
- A **stale** allowlist entry (naming a rule that no longer exists in
  `Deduce.lark`) also fails, so the allowlist can't silently rot.
- The success line now reports that all user-facing rules are documented.

No grammar or reference content changes — this is purely a CI-hardening guard.
It runs wherever the checker already runs (`make tests-tokens`,
`.github/workflows/test_deduce.yml`, `.github/workflows/static.yml`).

### Verification

- `python3 gh_pages/scripts/reference_grammar.py` →
  `Checked 178 reference grammar rule(s) and 48 operator-precedence entr(y/ies)
  against Deduce.lark; all 81 user-facing grammar rule(s) are documented.`
- `python3 gh_pages/scripts/keywords.py` passes (rest of `tests-tokens`).
- Negative tests (run manually): dropping a documented rule from the allowlist
  fails with the undocumented-rule message; adding a nonexistent rule to the
  allowlist fails with the stale-entry message; restoring passes.

ruff/mypy are not installed in this container, so they were not run locally;
the added lines are ≤83 chars and fully type-annotated, and CI runs both.

Resume this session on ginger: `~/deduce-runner/resume.sh d2fa7c50-b31c-4503-931f-55c947917e7f routine/20260807-180202`

### DIFF
--- a/gh_pages/doc/ImperativeReference.md
+++ b/gh_pages/doc/ImperativeReference.md
@@ -249,6 +249,26 @@ This is parser/AST only: proof-slot goal generation, ambiguity checking
 between theorem names and slot labels, and tying `h.valid_post` to a
 call label all land with the verifier in a later phase.
 
+## Allocation (Term)
+
+```deduce-grammar
+atomic_term ::= "new" IDENT new_type_args "(" term_list ")"
+new_type_args ::= ε
+new_type_args ::= "<" type_list ">"
+```
+
+`new Obj<T..>(args)` allocates a fresh instance of an
+[object](#object-statement) type `Obj`, passing the constructor arguments
+`args`. The optional angle-bracket type arguments instantiate a generic
+object's type parameters; unlike a procedure/object binding list these are
+*types*, so a compound argument such as `new Node<List<T>>(...)` parses.
+
+Allocation is only valid as the right-hand side of a `var`/assignment inside a
+procedure body (for example `var c := new Cell<Nat>(0)`); the `new` keyword in
+any other, pure-term position is a parse error. It requires
+`--experimental-imperative` and, like the other imperative forms, is currently
+accepted but not verified.
+
 ## Resource (Statement)
 
 ```deduce-grammar
--- a/gh_pages/scripts/reference_grammar.py
+++ b/gh_pages/scripts/reference_grammar.py
@@ -56,6 +56,25 @@ SUBSET_RULES = {
     "visibility",
 }
 
+# Deduce.lark rules that intentionally have no ``deduce-grammar`` fragment in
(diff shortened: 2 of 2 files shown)


jsiek:
Thanks @codex — good catch, and agreed. `new Obj<T..>(args)` is a real accepted surface form (an `atomic_term` alternative parsed into `ImpAlloc` in an imperative RHS), so allowlisting `new_type_args` as "internal" was papering over a genuine documentation gap: `ImperativeReference.md` documented `imp_rhs ::= term` but never the allocation form itself.

Rather than suppress the rule, I've **documented the allocation** and dropped `new_type_args` from the allowlist:

- New **"Allocation (Term)"** section in `gh_pages/doc/ImperativeReference.md` with a checked `deduce-grammar` block:
  ```
  atomic_term ::= "new" IDENT new_type_args "(" term_list ")"
  new_type_args ::= ε
  new_type_args ::= "<" type_list ">"
  ```
  plus prose covering the generic type args, the imperative-RHS-only restriction, and the `--experimental-imperative` / accepted-but-not-verified status.
- `new_type_args` removed from `INTERNAL_UNDOCUMENTED_RULES`.

The reverse guard now reports **82** user-facing rules documented (was 81). I verified against both parsers that the documented shapes parse in an imperative RHS — `new Cell<Nat>(0)`, the empty-type-args `new Cell(0)` (the `ε` alternative), and the compound `new Node<List<Nat>>(0)` — and that `new` in a pure-term position still errors, so the block matches the grammar it documents.

Issue #1166, opened 2026-08-07
array_bounds_goal builds a bare-name `<` that won't discharge against a resolved `<` premise


jsiek:
## Bug

`imperative_verifier.array_bounds_goal(read)` builds the bounds proof goal `i < length(a)` with a **bare-name** `ResolvedVar('<')` (mirroring the `mkEqual` idiom for `=`). But unlike `=` — which the checker/`reduce` treat specially everywhere — `<` is an ordinary stdlib function, so a bare-name `<` does **not** reduce to (or compare equal to) the *resolved/uniquified* `<` that a user's `requires i < length(a)` premise carries. When the read-bounds path is eventually wired to discharge these obligations, they will never discharge against a real premise of the same surface shape.

The code comment claims it "matches a `requires i < length(a)` precondition of the same shape once `discharge` reduces both sides" — that claim holds for `=` but is false for `<`.

## Repro

Confirmed while implementing #1118 (Phase 2i). Discharging a bounds obligation built by `array_bounds_goal` against a matching `requires i < length(a)` premise fails; with `--unique-names` the goal shows a bare `<` while the premise shows `<.s1_s1_s3_s13_0`:

```
KIND array bounds
GOAL i < length(a)
ANTE i <.s1_s1_s3_s13_0 length(a)
```

`check_implies` then reports the bounds obligation as an incomplete proof even though the premise is literally the same inequality.

The obligation is not yet wired to a body-read-walking pass (#1117 deferred that), so no current test exercises discharge — which is why this was latent.

## Observed vs. expected

- **Observed:** a bounds obligation `i < length(a)` built by `array_bounds_goal` cannot be discharged from a `requires i < length(a)` premise, because the two `<` operators differ (bare-name vs resolved).
- **Expected:** the obligation discharges from a matching premise.

## Fix already available

For mutable-array *writes*, #1118 sidesteps this in `checker_pipeline._array_bounds_obligation`, which builds the `<`/`length` call and runs it through `type_check_formula` so the operators resolve against their operand types exactly like the user's premise. The read-collector path in `imperative_verifier` should adopt the same resolved-operator construction (it needs access to the environment) when it is wired for discharge.

Refs #854, #1117, #1118.

jsiek:
[...]

Issue #1119, opened 2026-07-27
Phase 2j: Enforce procedure and loop modifies frames for arrays


jsiek:
Parent: #854

Phase: 2 dynamic-frame enforcement for mutable arrays.

## Goal

Reject every mutable-array write that is not covered by the active procedure and loop modifies frames.

## Scope

- Normalize Phase 2 frame expressions for whole arrays, single cells, comma unions, and `{}`.
- Treat an omitted procedure `modifies` clause as the empty frame.
- Check each array write against the procedure frame.
- When inside a loop, also check against the loop frame; define the default loop frame explicitly.
- Use existing arithmetic/proof implication machinery to prove cell inclusion when indices are symbolic.
- Produce the requested diagnostic showing the written location and active frame.

## Acceptance

- [ ] Whole-array frames permit any in-bounds cell write.
- [ ] `modifies a[i]` permits that cell and rejects an unrelated cell.
- [ ] Empty/omitted frames reject writes.
- [ ] Nested loop and procedure frames are both enforced.
- [ ] Add `array_swap_missing_modifies.pf` as a golden error fixture.

## Dependencies

Depends on #1111 and #1118.

## Parallelization

Can run in parallel with entry-state `old` support (#1120). It unlocks array swap and loop examples.

## Out of scope

No object fields, footprint expressions, allocation freshness, or disjointness inference beyond array/cell frames.

PR #1167, opened 2026-08-07
Phase 2i: model mutable-array writes in the verifier (closes #1118)

### LINKED ISSUE #1118: Phase 2i: Model mutable-array writes in symbolic verifier state
Parent: #854

Phase: 2 mutable-array write support.

## Goal

Give `a[i] := v` a verifier-only state transition that subsequent expressions and postconditions can reason about.

## Scope

- Introduce the minimal symbolic heap/state representation needed for mutable arrays.
- Type-check the array handle, index, and assigned value.
- Reuse the bounds-obligation path from #1117 for writes.
- Model a write as a fresh array-state version or equivalent functional update.
- Define and test read-after-write behavior for equal and unequal indices.
- Preserve immutable array length across updates.
- Keep state representation internal; do not expose generated names in user diagnostics.

## Acceptance

- [ ] `a[i] := v; assert a[i] = v` verifies with an in-bounds premise.
- [ ] A write of the wrong element type is rejected.
- [ ] Reads at a provably different index retain the previous value.
- [ ] Repeated writes compose in source order.
- [ ] Unit tests pin symbolic update and read-over-write formulas.

## Dependencies

Depends on #1114, #1115, and #1117.

## Parallelization

After this lands, modifies-frame enforcement and `old` snapshots can proceed in parallel.

## Out of scope

No frame permission check, allocation, object fields, or runtime mutation.

jsiek:
## Summary

Phase 2i of the imperative verification layer (#854): give a mutable-array
element write `a[i] := v` a verifier-only state transition so subsequent reads
and postconditions reason about it by read-over-write.

Closes #1118. Rebased onto `main` after 2g (#1165) landed; the write-modeling is
folded into 2g's recursive `walk`.

### Core

- **New AST node `ArraySet(subject, position, value)`** (`abstract_syntax/terms.py`)
  — the array `subject` with index `position` updated to `value`. Verifier-only
  (never parsed). `ArrayGet.reduce` does read-over-write against it (written
  value at the written index — even symbolic; prior value at a provably-different
  *concrete* index; symbolic otherwise, so a symbolic `i ≠ j` never reads through
  — sound). `ArrayLength.reduce` sees through it (a write preserves length).
  Added to all four `rewrite.py` mark walkers and the AST specimen factories.
- **`checker_pipeline.py`**: type-check `a[i] := v` and model it in the `walk`
  as a functional `ArraySet` update of the array's symbolic state; emit the
  write's in-bounds obligation `i < length(a)`.

### Soundness (Codex review of the first cut — both P1s were real)

- **Aliasing.** Keying write-state by handle *name* is sound only when handles
  can't alias. The design (`docs/imperative-verification-plan.md`, §"One heap
  model") resolves aliasing with an **identity-keyed heap + dynamic frames** —
  both later phases (frames = #1119). So `_array_write_aliasing_risk` **defers**
  any *writing* proc with more than one reachable mutable-array handle (>1
  `[T]!` param, or a handle-copy local). Single-array procs (all the #1118
  acceptance cases) stay verifiable and lift unchanged when the heap lands;
  `ArraySet` is that heap's inner (element) store.
- **Read bounds.** Every executed read `a[j]` now raises `j < length(a)` (the
  design's "all array accesses are in bounds" goal) via a generic `ArrayGet`
  walk. Writes already emitted their bounds.
- **Exit-state postconditions.** `ensures` clauses are now substituted with the
  write state, fixing an unsoundness where a postcondition over a written array
  ignored the write.
- Bounds goals are built by type-checking a `<`/`length` call so the operators
  resolve and match a `requires i < length(a)` premise (bare-name `<` does not;
  #1166).

### Acceptance (issue #1118)

`a[i]:=v; assert a[i]=v` verifies with an in-bounds premise ✅ · wrong element
type rejected ✅ · read at a provably-different concrete index keeps the prior
value ✅ · repeated writes compose ✅ · unit tests pin the read-over-write
formulas ✅ · length preserved across writes ✅.

### Tests

`read_demo` (#1117) now verifies → moved to should-validate; new
should-validate/imperative_array_write.pf; should-error write/read-bounds +
element-type fixtures; should-warn aliasing-deferral regression;
test/unit/test_array_set.py. Full `test-deduce.py` (both parsers), unit suites,
`ruff`, `mypy` all green.

Resume this session on ginger: `~/deduce-runner/resume.sh a2b1e949-a20e-45dc-8bd8-c9a553015523 routine/20260807-230202`


jsiek:
Thanks for the review — both P1s were real soundness bugs, and I've reworked the slice to fix them. I also rebased onto `main` now that 2g (#1165) landed, folding the write-modeling into its recursive `walk`.

## Aliasing (P1, `checker_pipeline.py:565`)

This is the important one, and I grounded the fix in the design doc (`docs/imperative-verification-plan.md`, §"One heap model"). The intended state is `heap + local variables + allocation set`, where the heap maps **identity-keyed** locations `array_cell(array_id, index)` to values. A `[T]!` handle carries an *array id*; two handles with the same id are the same array, so a write keyed by id is visible through every alias, and `new` picks a *fresh* id. Aliasing is thus resolved by **array identity + dynamic frames** (`reads`/`modifies`/`old`/`footprint`) — and both the identity-keyed heap and frames are *later* phases than this one (frames are #1119).

My per-**name** `ArraySet` is the correct *inner (element)* store of that heap, but the outer identity/frame layer that disambiguates aliases doesn't exist yet. So for #1118 I keep the verifier inside the fragment where aliasing **cannot** arise: `_array_write_aliasing_risk` now **defers** any *writing* procedure with more than one reachable mutable-array handle (>1 `[T]!` parameter, or a handle-copy local `var y := a`). With a single handle there is nothing to alias with, so the per-name update is exactly sound, and single-array procs lift unchanged once the identity heap arrives. Your exact counterexample now defers (emits the "not verified" warning) instead of falsely verifying — pinned by `test/should-warn/imperative_array_alias.pf`.

## Read bounds (P1, `checker_pipeline.py:441`)

Fixed. Every executed read `a[j]` now raises an in-bounds obligation `j < length(a)` (the design's "all array accesses are in bounds" goal), collected by a generic `ArrayGet` walk (`_collect_array_gets`) over each executed term — `var`/assignment RHS, write index/value, `if` condition, `assert`, and `return` (not `assume`, which is trusted like `requires`). `return a[i]` with no premise is now rejected — pinned by `test/should-error/imperative_array_read_unbounded.pf`.

## Also fixed in this pass

- **Postconditions weren't evaluated in the exit state.** An `ensures` mentioning a written array ignored the write, so a false postcondition could verify; `_proc_postconditions` now substitutes the write state.
- Bounds goals are built by *type-checking* a `<`/`length` call so the operators resolve and match a `requires i < length(a)` premise. The bare-name `<` in the still-unwired read-collector (`imperative_verifier.array_bounds_goal`) does **not** discharge this way — filed as #1166.

Full `test-deduce.py` (both parsers), the unit suites, `ruff`, and `mypy` are green.

Issue #1133, opened 2026-07-27
Phase 2x: Close out Phase 2 documentation, diagnostics, and regression coverage


jsiek:
Parent: #854

Depends on: #1106, #1109, #1110, #1124, #1125, #1127, #1128, #1129, #1130, #1131, #1132

## Goal

Perform the release-quality pass for the Phase 1 and Phase 2 implementation before #854 is marked complete.

## Work

- Update the imperative plan and language reference with the exact implemented grammar, verifier semantics, limitations, and examples.
- Replace the temporary unchecked-declaration warning from #1108 for supported procedures and observers; retain clear diagnostics for constructs still accepted only syntactically.
- Add or stabilize diagnostic assertions for representative parse, type, frame, proof, invariant, decreases, label, bounds, and postcondition failures.
- Build a coverage matrix mapping every Phase 2 feature and acceptance example to at least one passing and one failing regression where meaningful.
- Confirm the normal CI workflow runs the imperative CLI suite and the imperative LSP symbol regressions.
- Run all repository test suites and record any unrelated pre-existing failures in this issue.
- Update #854 with the final checklist and close it only after every required dependency is complete.

## Acceptance criteria

- local_counter, array_swap, array_fill_zero, linear_search, frame_preserves_sorted, binary_search, procedure-call, proof-slot, and allocation fixtures are all covered in CI.
- Both parser implementations are covered for the supported imperative syntax.
- The documentation no longer claims either that nothing is implemented or that verified declarations are merely ignored.
- Unsupported object/resource semantics and runtime execution are explicitly identified as later work.
- #854 has an auditable completion comment linking the test evidence.

## Out of scope

- Phase 3 ownership/resource verification.
- Runtime interpretation or compilation of procedures.
- New language features discovered during the closeout; file those separately.

PR #1173, opened 2026-08-08
Phase 2n: verify out-of-line proof slots and by-clauses (closes #1123)

### LINKED ISSUE #1123: Phase 2n: Check inline proofs and out-of-line imperative proof slots
Parent: #854

Phase: 2 proof-slot verification.

## Goal

Connect imperative `by` clauses and procedure proof blocks to the exact source-located verification obligations they discharge.

## Scope

- Check ordinary inline proof expressions against attached obligations.
- Treat `by label` as a local proof-slot declaration/reference and match it to exactly one proof-block entry.
- Reject missing, duplicate, unused, and unknown proof-block entries.
- Detect ambiguity when a bare name could mean both an in-scope theorem and a local slot.
- Check proof-block bodies in the givens/context captured at the source annotation, not at the end of the procedure.
- Support `assert`, loop `established`, loop `preserved`, and `decreases ... by` attachments as their obligation kinds become available.

## Acceptance

- [ ] Inline `by .` and theorem references are checked, not merely parsed.
- [ ] A long proof moved to `proof ... end` sees the same goal/givens as the inline form.
- [ ] Missing, duplicate, unused, unknown, and ambiguous labels have focused fixtures.
- [ ] Proof slot names do not leak into module exports.
- [ ] Existing non-imperative proof checking remains unchanged.

## Dependencies

Depends on #1112 and #1115.

## Parallelization

Can run in parallel with loop, array, and call semantics. Later slices only need to attach their obligations to this API.

## Out of scope

No call-postcondition labels; those are #1125. No new automation.

jsiek:
## Phase 2n: verify out-of-line proof slots and `by`-clauses

Closes #1123. Phase 2n of the imperative verification layer (issue #854).

Connects imperative `by` clauses and a procedure's out-of-line
`proof ... end` block to the exact source-located obligations they discharge.

### What changed

- **`verify_proc` no longer defers a procedure just because it declares a
  proof block.** A `by <slot>` clause resolves (via uniquify's pre-bound slot
  labels) to a bare `PVar` naming one `proof ... end` entry. The verifier
  matches it to that entry and *inlines* the entry's proof into the obligation
  it discharges, so a moved-out proof is checked against the **very same goal
  and givens the inline form sees** — captured at the source annotation, not at
  the end of the procedure. A fact established later in the body is therefore
  not in scope for an earlier slot's proof. In-scope `requires`/`assume`/`assert`
  facts remain reachable by formula via `recall (<fact>)`.

- **Malformed proof blocks are rejected with focused diagnostics:**
  - `duplicate` — two entries share a label (rejected in uniquify);
  - `ambiguous` — a slot label also names an in-scope theorem, so a bare
    `by <label>` could mean either (rejected in uniquify with a clear
    proof-slot message, replacing the generic "cannot overload" error);
  - `unused` — an entry no `by` clause cites (rejected in `verify_proc`);
  - `unknown`/`missing` — a `by <label>` naming neither a slot nor a theorem is
    the standard `undefined proof variable` diagnostic (uniquify). These two
    acceptance items surface as the same diagnostic and are covered by one
    fixture.

- **Proof-slot labels stay local to their procedure and never become module
  exports** (`proc_proof_slot_no_export.pf` pins this).

### Acceptance criteria (#1123)

- [x] Inline `by .` and theorem references are checked, not merely parsed
      (`proc_assert_by_theorem.pf`, `proc_proof_slot_false.pf`).
- [x] A long proof moved to `proof ... end` sees the same goal/givens as the
      inline form (`proc_proof_slot.pf`: a `requires` given used via `recall`
      plus a multi-step proof moved out of line; guaranteed by inlining).
- [x] Missing, duplicate, unused, unknown, and ambiguous labels have focused
      fixtures (`proc_proof_slot_{unknown,duplicate,unused,ambiguous}.pf`, plus
      `late_fact` for context capture).
- [x] Proof slot names do not leak into module exports
      (`proc_proof_slot_no_export.pf`).
- [x] Existing non-imperative proof checking remains unchanged (full suite).

### Testing

- `python3 test-deduce.py --imperative` — all imperative fixtures pass under
  both parsers (RD and LALR), including the 8 new should-error and 2 new
  should-validate fixtures.
- `python3 test-deduce.py` (CI-equivalent: lib, should-validate, should-error,
  should-warn, prelude, parser equivalence) — all pass.
- `ruff` + `mypy` clean on the changed source files.
- Unit tests in `test/unit/test_imperative_verify.py` (21 pass), including new
  coverage for slot inlining and the unused-slot error.

Note: the `--imperative` lane is not yet part of the CI matrix (that is #1133's
closeout work, and this run cannot push workflow changes); the new fixtures
follow the same in-tree convention as the earlier Phase 2 PRs (#1160, #1165).
The uniquify changes still get CI coverage via the `proc` fixtures in
`EXPERIMENTAL_IMPERATIVE_FILES` that run under `--passable`/`--errors`.

Resume this session on ginger: `~/deduce-runner/resume.sh 94a0a896-e201-4732-adb8-5719ee917bac routine/20260808-040202`


jsiek:
Good catch — this was a real gap, now fixed.

The original ambiguity check ran in `ProcDecl.uniquify` against `body_env['no overload']`, which is only populated by **locally declared** theorems/functions. An imported theorem reaches the uniquify env as an ordinary proof-variable binding but never enters `no overload`, so the check missed it and the slot pre-binding silently shadowed the import — exactly as described. Repro before the fix:

```
import Base                       // Base declares `theorem ex_mid: all b:bool. b or not b`
proc p(x: bool) {
  assert x = x by ex_mid
}
proof
  ex_mid { . }                    // was accepted; `by ex_mid` resolved to the slot
end
```

**Fix:** the ambiguity check moved out of uniquify and into `verify_proc` (`_reject_ambiguous_proof_slots`), where it runs against the **typed** proof environment. Both local and imported theorems are `ProofBinding` entries there, so a slot label matching either (by base name) is rejected uniformly; parameters/locals are `TermBinding`s and are correctly not flagged. The check runs for every `proc` — verifiable or deferred — since the clash is purely a naming question. Uniquify now binds slot labels directly (keeping only the duplicate-label check).

Regression: `test/should-error/imperative_proof_slot_imported_ambiguous.pf` (imports `Base`, names a slot `ex_mid`), enrolled in `EXPERIMENTAL_IMPERATIVE_FILES` so it runs in the CI `--errors` leg. The existing local-theorem fixture (`proc_proof_slot_ambiguous.pf`) still reports the same location and message.

Issue #1183, opened 2026-08-11
The website code highlighter is presently borked


EmWue:
The code examples on the website are presently just littered with stray `<span class="operator">`s, seemingly causes by the opening `<` (note the different highlighting color on this character) being for some reason double-replaced. It only seems to happen when the class assigned is `operator`, but i haven't looked through every code example to check (I think none of the code examples contain comments, perhaps worth checking if they are affected too).

I would wager the culprit is somewhere in or near the `// (heavy quote) lexing (heavy quote)` section of the `codeToHtml` function in `codeUtils.js`, or alternatively some change made to the `codeKeywords.js` file that had unintended consequences, but I didn't spot any obvious mistakes.

Steps to reproduce:
- Look at any of the code examples on [the deduce website](https://jsiek.github.io/deduce/)
- Perceive the borked unreadability of them

EmWue:
I did some debugging, and while I still don't fully understand the problem (as in, no idea why it doesn't trigger in even more places), the problematic step seems to be applying the keywords regex `kre`. And upon some inspection.... `keywords` in the `codeKeywords.js` files contains the following regex fragment:
`[\-\+∸*%=≠<>≤≥&∘∪∩⊆∈⨄⊝\\^≲≈]`
Note that this has some overlap with what is defined in `operators` (probably need to be fixed, but not the root cause here), but it also neglects to escape certain symbols as their HTML character entity equivalents.
Simply yeeting `<` and `>` from keywords appears to provide a temporary fix, but it's more of a hack than a real solution.

PR #1184, opened 2026-08-13
Unbork the website by changing how the js regexes are autogenerated

EmWue:
A previous commit removed the rewriting of certain characters into HTML entities inside regex character classes, since those do not understand HTML entities, fixing some syntax highlighting issues.

Regrettably, not rewriting characters inside character classes in turn causes issue #1183, and the fix is to rewrite the character classes by first removing the offending characters from the class, and then adding the corresponding HTML entity as an alternative at the end of the class.


jsiek:
Thank you!

